PR #4, opened 2015-01-27
Fix the EPoll selector and the FutureTask to be much better

xjdr:
This will allow us to Modify the Selector, Future and FutureTask for performance tuning in the future.


pdex:
I have no idea what you changed here, can we break this out into two PRs so we can keep track of how we're diverging from the JDK source?

pdex:
I'll take that as a no. :)

PR #5, opened 2015-01-29
Rest support

xjdr:
Added main skeleton for basic REST support.  
Made some significant improvments on heading off the long tail in the HttpParser.


xjdr:
It's functional so I am going to proceed to merge.

PR #37, opened 2016-04-14
Vip jr fix

xjdr:


src/main/java/com/xjeffrose/xio/client/loadbalancer/Distributor.java
```diff
@@ -84,7 +83,6 @@ public Node getNodeById(UUID id) {
    */
   public Node pick() {
     if (revLookup.size() < 1) {
```
xunlai: NPE can still happen in this public pick() method.

I think here the content of this method can just be replaced by one line
return pick(0);
or 
return pick(1)

PR #39, opened 2016-04-14
Vip jr fix oob

xjdr:


xunlai:
This is for master branch.  Closing now

PR #56, opened 2016-04-28
Remove redundant HttpClient DisconnectHandler

otoolec:
Also fixes the confusion caused by milli vs sec

### DIFF
--- a/src/main/java/com/xjeffrose/xio/client/HttpClientConnector.java
+++ b/src/main/java/com/xjeffrose/xio/client/HttpClientConnector.java
@@ -2,10 +2,7 @@ package com.xjeffrose.xio.client;
 
 
 import com.google.common.net.HostAndPort;
-import com.xjeffrose.xio.core.ChannelStatistics;
-import com.xjeffrose.xio.core.ConnectionContextHandler;
-import com.xjeffrose.xio.core.XioExceptionLogger;
-import com.xjeffrose.xio.core.XioSecurityHandlers;
+import com.xjeffrose.xio.core.*;
 import com.xjeffrose.xio.server.IdleDisconnectHandler;
 import io.netty.channel.Channel;
 import io.netty.channel.ChannelInitializer;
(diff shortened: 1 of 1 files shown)


src/main/java/com/xjeffrose/xio/client/HttpClientConnector.java
```diff
@@ -101,15 +98,11 @@ protected void initChannel(SocketChannel channel) throws Exception {
...
-            20000,
-            NO_WRITER_IDLE_TIMEOUT,
-            NO_ALL_IDLE_TIMEOUT,
-            TimeUnit.MILLISECONDS));
-        cp.addLast("idleDisconnectHandler", new IdleDisconnectHandler(
-            2000,
-            NO_WRITER_IDLE_TIMEOUT,
-            NO_ALL_IDLE_TIMEOUT));
+        cp.addLast("idleDisconnectHandler", new XioIdleDisconnectHandler(
+              20000,
```
otoolec: I'm actually not sure if this is the right timeout.  2,000 seconds seemed like a long time.

xjdr: updated in master with the following 

```java
       if (def.getClientIdleTimeout() != null) {
          cp.addLast("idleDisconnectHandler", new XioIdleDisconnectHandler(
              (int) def.getClientIdleTimeout().toMillis(),
              NO_WRITER_IDLE_TIMEOUT,
              NO_ALL_IDLE_TIMEOUT,
              TimeUnit.MILLISECONDS));
        }
```

otoolec: I only see that having been done for [XioServerTransport](https://github.com/xjdr/xio/blob/master/src/main/java/com/xjeffrose/xio/server/XioServerTransport.java#L91).  This PR is for doing a similar change in HttpClientConnector.

Should XioClientConfig be modified to add a client idle timeout?

xjdr:
Sorry, misread the filename. Merged

src/main/java/com/xjeffrose/xio/client/HttpClientConnector.java
xjdr: Yes, it probably should be. I'm not a big fan of static timeout values.

PR #57, opened 2016-04-29
always close the health check channel

xunlai:


src/main/java/com/xjeffrose/xio/client/loadbalancer/Node.java
```diff
@@ -112,25 +112,25 @@ public boolean isAvailable() {
...
-    } catch (IOException e) {
       try {
-        retryLoop.takeException(e);
-      } catch (Exception e1) {
-        return false;
+        if (!connectionStopwatch.isRunning()) {
+          connectionStopwatch.start();
+        }
+        try (SocketChannel channel = SocketChannel.open()) {
+          channel.connect(address);
```
smarth-madan: Where is the channel.close() for this ?

PR #101, opened 2016-08-09
Fix Node write failure case

otoolec:
Need to use channelFuture.cause() if the writeAndFlush fails
for Node.send since future.cause() will alwasy be empty
when writeAndFlush is being called.


pdex:
Good catch.

PR #102, opened 2016-08-09
Make AsyncRetryLoop thread safe

otoolec:
Needs to be thread safe since it uses EventLoopGroup instead
of EventLoop to schedule the retry

### DIFF
--- a/src/main/java/com/xjeffrose/xio/client/asyncretry/AsyncRetryLoop.java
+++ b/src/main/java/com/xjeffrose/xio/client/asyncretry/AsyncRetryLoop.java
@@ -1,11 +1,8 @@
 package com.xjeffrose.xio.client.asyncretry;
 
 import io.netty.channel.EventLoopGroup;
-import io.netty.channel.ConnectTimeoutException;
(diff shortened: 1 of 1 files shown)


pdex:
We should probably change this to use EventLoop. There shouldn't be more than one thread involved with a single instance of a given retry loop.

Additionally this was based off the retry logic in zookeeper, and as a result it has some leaky abstractions:

https://github.com/xjdr/xio/blob/develop/src/main/java/com/xjeffrose/xio/client/XioConnectionPool.java#L89

So if you're using this, it's probably going to change at some point in the future.

otoolec:
Will the XioConnectionPool be limited to a single EventLoop?  I'm still trying to wrap my head around how the threads from XioConnectionPool, AsyncRetryLoop, and the caller of Node.send will interact, or if we can guarantee that they are all the same thread. 

It seems like it should be possible to have everything in the same thread as long as the MultiNodeClient is constructed using a bootstrap that only contains the callers EventLoop for the EventLoopGroup, but I don't see how we can guarantee that.

PR #104, opened 2016-08-11
Refactor XioServerConfig to use lombok.Getter

pdex:


src/main/java/com/xjeffrose/xio/server/XioServerConfig.java
```diff
@@ -5,112 +5,40 @@
...
 import io.netty.util.Timer;
+import lombok.Getter;
 
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
 import java.net.InetSocketAddress;
 
 public class XioServerConfig {
-  // old
+  @Getter
```
otoolec: Could a class level @Getter annotation be used here?  Is that being avoided so that any future fields don't accidentally get exposed with a getter?

pdex: I'm being pedantic here, someone who isn't familiar with lombok might be surprised by the class level @Getter

PR #137, opened 2017-06-28
Cleanup random scraps and add notes

pdex:


Notes.org
```diff
@@ -0,0 +1,51 @@
...
+
+** clean up javadoc linting errors
+
+** add http Recipes
+
+* speeding up maven invocation
+
+** drip
+
+*** https://github.com/ninjudd/drip
```
xjdr: we should also look at nailgun (vs drip)

pdex: Nailgun is old, drip is new. Honestly I'm not seeing terrible JVM startup performance on my laptop so the point is moot.

PR #141, opened 2017-09-08
Add UrlRoundRobinRouter and allow set/get in ApplicationState

andyday:
using these changes in taco config-spike to dynamically update routes.


andyday:
This PR changes the following:

- Adds UrlRoundRobinRouter abstracted by HttpRouter.  This class combines the UrlRouter and RoundRobinRouter for balancing among hosts based on a url pattern.
- Fixed maven javadoc issues.
- Added set/get of HttpRouter in ApplicationState.  Http1ProxyFragment defaults to use appState.getHttpRouter if non is configured.

PR #144, opened 2017-10-13
Allow Url routing to a Round Robin balanced set of handlers.

andyday:
This PR changes the following:

- Renames RouteProvider to RequestHandler (based on prior TODO).
- Adds RoundRobinRequestHandler. 
- Abstracts UrlRouter with Router interface.
- Fixed maven javadoc issues.
- Added set/get of Router in ApplicationState.  Http1ProxyFragment defaults to use appState.getHttpRouter if non is configured.


src/main/java/com/xjeffrose/xio/client/loadbalancer/Distributor.java
```diff
@@ -74,34 +70,46 @@ public void stop() {
 
   /**
    * The vector of pool over which we are currently balancing.
+   *
+   * @return Node list
```
pdex: We should probably find a tool to generate javadocs so we can make future diffs more readable.

example-trailhead/src/main/java/com/xjeffrose/xio/server/trailhead/Trailhead.java
```diff
@@ -26,22 +20,22 @@
   private final ApplicationBootstrap bootstrap;
   private Application application;
 
-  private RouteProvider buildProvider(ProxyConfig proxyConfig) {
-    return new SimpleProxyRoute(proxyConfig);
+  private RequestHandler buildProvider(Route route, ProxyConfig proxyConfig) {
```
pdex: I waffled on calling this RequestHandler in the past. I don't have a good rubric for how we should name these things, we should probably spend some time clarifying concepts post PoC.

src/main/java/com/xjeffrose/xio/client/loadbalancer/Distributor.java
```diff
@@ -32,11 +32,7 @@
   private final Timeout refreshTimeout;
 
   private final Ordering<Node> byWeight = Ordering.natural().onResultOf(
-      new Function<Node, Integer>() {
-        public Integer apply(Node node) {
-          return node.getWeight();
-        }
-      }
+    (Function<Node, Integer>) node -> node.getWeight()
```
pdex: Did intellij generate this cast? You should be able to get away with 
```java
private final Ordering<Node> byWeight = Ordering.natural()
                                                .onResultOf((node) -> node.getWeight())
                                                .reverse();
```

src/main/java/com/xjeffrose/xio/http/Http1ProxyFragment.java
```diff
@@ -30,7 +34,7 @@ public String applicationProtocol() {
   }
 
   public void buildHandlers(ApplicationState appState, XioServerConfig config, XioServerState state, ChannelPipeline pipeline) {
-    pipeline.addLast(new Http1ProxyHandler(router));
+    pipeline.addLast(new Http1ProxyHandler(router != null ? router : appState.getRouter()));
```
pdex: If we're pulling the router in from the appState we should remove the router member here and the corresponding constructor.

src/main/java/com/xjeffrose/xio/http/Http1ProxyHandler.java
```diff
@@ -11,16 +11,20 @@
...
+  public Http1ProxyHandler(Router router) {
     this.router = router;
   }
 
   @Override
   public final void channelRead0(final ChannelHandlerContext ctx, HttpObject msg) throws Exception {
+    if (router == null) {
+      // Router is null meaning it has not been set yet.
+      throw new IllegalStateException("No Router");
+    }
```
pdex: Don't do this. ApplicationState.getRouter should be able to provide this handler a router at pipeline construction time. If you absolutely cannot get a router configuration, instantiate a NullRouter instance. I'm of the opinion that there should be a sane default router configuration that we can load at startup.

src/main/java/com/xjeffrose/xio/http/SimpleProxyHandler.java
```diff
@@ -6,86 +6,81 @@
...
-    Optional<String> path = route.groups(payload.getUri())
+    Optional<String> path = route.groups(payload.uri())
       .entrySet()
       .stream()
       .filter(e -> e.getKey().equals("path"))
       .map(e -> e.getValue())
       .findFirst();
 
-    payload.setUri(path.map(config.urlPath::concat).orElse(config.urlPath));
+    if (!config.pathPassthru) {
```
pdex: Are we expecting a service to respond to the full url 

`https://api.nordstrom.com/v1/backend/foo`

instead of 

`/foo`

?

My preference is to simplify things for the backend service as much as possible

src/main/java/com/xjeffrose/xio/client/loadbalancer/Distributor.java
andyday: ya.  good ole intellij.  and i only touched that file to fix the javadoc errors...

example-trailhead/src/main/java/com/xjeffrose/xio/server/trailhead/Trailhead.java
andyday: i saw the todo and thought requesthandler made more sense to me so i went for it...  not totally married to it or anything though.

src/main/java/com/xjeffrose/xio/client/loadbalancer/Distributor.java
andyday: ya.  i normally make format/style changes in separate prs, but this one got away with me.  what's weird is it only gives errors if you've started creating a javadoc but it is invalid.  doesn't say anything about stuff that has no javadoc comments.  i've found that javadoc comments are only useful if they provide context and maybe examples of usage...

src/main/java/com/xjeffrose/xio/http/Http1ProxyFragment.java
andyday: agreed.  will rework.

src/main/java/com/xjeffrose/xio/http/Http1ProxyHandler.java
andyday: i can totally create a default or null router for this.   it minimum it should at least log a warning so that the logs tell the tale.   as for sane default router config, it would be hard because all this comes from the registry and we can make no assumptions about what that config might be...   for now i'll create a default router and just log usage...

src/main/java/com/xjeffrose/xio/http/SimpleProxyHandler.java
andyday: this setting is for nfe --> nlp.  in this case we want the entire original url to pass through to the nlp because it uses the prefix same as nfe does.  nlp then does the path replacement as usual.

andyday: agreed on keep it simple for backend service.

src/main/java/com/xjeffrose/xio/client/loadbalancer/Distributor.java
andyday: this didn't work.  reverted back to my original

src/main/java/com/xjeffrose/xio/http/SimpleProxyHandler.java
andyday: fixed this.  added default router and eliminated the null case.

PR #145, opened 2017-10-19
Added Mutual Auth

pdex:


src/main/java/com/xjeffrose/xio/core/XioMessageLogger.java
```diff
@@ -6,6 +6,8 @@
 import io.netty.handler.logging.LoggingHandler;
 import lombok.extern.slf4j.Slf4j;
 
+//TODO(CK): Refactor this logMessageDebug is redundant
+//TODO(CK): Find a use case for this, why not just LoggingHandler?
```
xjdr: This does the fancy byte level debug logic. Very useful for protocol level debugging for custom codecs

pdex: Take a closer look, we're not using any of that. :)

Issue #156, opened 2017-12-16
wrap http2 objects


pdex:
https://github.com/xjdr/xio/pull/152/files#diff-b14238ce9f35c5bb385d1c330684df66R39

#152

PR #166, opened 2017-12-27
Wrap Http2 objects

pdex:
Addresses #156 

* [x] Wrap h2 Headers
* [x] Wrap h2 FullRequest
* [x] Wrap h2 StreamingRequest
* [x] Wrap h2 FullResponse
* [x] Wrap h2 StreamingResponse
* [x] Http2ServerCodec
* [x] Http2ClientCodec

### DIFF
--- a/src/main/java/com/xjeffrose/xio/http/Http2ClientCodec.java
+++ b/src/main/java/com/xjeffrose/xio/http/Http2ClientCodec.java
@@ -1,0 +1,137 @@
+package com.xjeffrose.xio.http;
+
+import com.xjeffrose.xio.core.internal.UnstableApi;
+import com.xjeffrose.xio.http.internal.FullHttp2Response;
+import com.xjeffrose.xio.http.internal.Http2StreamingData;
+import com.xjeffrose.xio.http.internal.StreamingHttp2Response;
+import io.netty.channel.ChannelDuplexHandler;
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelPromise;
+import io.netty.handler.codec.http2.DefaultHttp2DataFrame;
+import io.netty.handler.codec.http2.DefaultHttp2Headers;
+import io.netty.handler.codec.http2.Http2DataFrame;
+import io.netty.handler.codec.http2.Http2Headers;
+import io.netty.util.AttributeKey;
+import io.netty.util.concurrent.PromiseCombiner;
+import lombok.extern.slf4j.Slf4j;
+
+@UnstableApi
+@Slf4j
+public class Http2ClientCodec extends ChannelDuplexHandler {
+
+  private static final AttributeKey<Response> CHANNEL_RESPONSE_KEY =
+      AttributeKey.newInstance("xio_channel_h2_response");
+
+  private static void setChannelResponse(ChannelHandlerContext ctx, Response response) {
+    ctx.channel().attr(CHANNEL_RESPONSE_KEY).set(response);
+  }
+
+  private static Response getChannelResponse(ChannelHandlerContext ctx) {
+    // TODO(CK): Deal with null?
+    return ctx.channel().attr(CHANNEL_RESPONSE_KEY).get();
+  }
+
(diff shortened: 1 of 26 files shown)


src/main/java/com/xjeffrose/xio/http/Http2ClientCodec.java
```diff
@@ -0,0 +1,137 @@
...
+  private static final AttributeKey<Response> CHANNEL_RESPONSE_KEY =
+      AttributeKey.newInstance("xio_channel_h2_response");
+
+  private static void setChannelResponse(ChannelHandlerContext ctx, Response response) {
+    ctx.channel().attr(CHANNEL_RESPONSE_KEY).set(response);
+  }
+
+  private static Response getChannelResponse(ChannelHandlerContext ctx) {
+    // TODO(CK): Deal with null?
+    return ctx.channel().attr(CHANNEL_RESPONSE_KEY).get();
```
xjdr: I'd prefer to wrap the get in an `Optional<>` and handle the `ifPresent()` conditions

src/main/java/com/xjeffrose/xio/http/Http2ClientCodec.java
```diff
@@ -0,0 +1,137 @@
...
+      }
+    } else if (msg.payload instanceof Http2DataFrame) {
+      Response response =
+          new StreamingResponseData(
+              getChannelResponse(ctx),
+              new Http2StreamingData(((Http2DataFrame) msg.payload).content(), msg.eos));
+      return response;
+    }
+    // TODO(CK): throw an exception?
+    return null;
```
xjdr: If we throw in the pipeline, we need to do a better job of doing something more meaningful with the exception handler

src/main/java/com/xjeffrose/xio/http/Http2ServerCodec.java
```diff
@@ -0,0 +1,137 @@
...
+  private static final AttributeKey<Request> CHANNEL_REQUEST_KEY =
+      AttributeKey.newInstance("xio_channel_h2_request");
+
+  private static void setChannelRequest(ChannelHandlerContext ctx, Request request) {
+    ctx.channel().attr(CHANNEL_REQUEST_KEY).set(request);
+  }
+
+  private static Request getChannelRequest(ChannelHandlerContext ctx) {
+    // TODO(CK): Deal with null?
+    return ctx.channel().attr(CHANNEL_REQUEST_KEY).get();
```
xjdr: See above re `Optional<>`

src/main/java/com/xjeffrose/xio/http/Http2ClientCodec.java
pdex: Well you get a null by default with Attribute :)

IMHO ifPresent doesn't solve the problem of it not being present when we expect it. So now we've introduced a bug that's the opposite of not dealing with null.

We're not exposing the result of getChannelResponse to the outside world here. I think we should focus on using Optional when we're presenting a potentially empty result to an xio user. I'm not sure it's a good fit for internals like this.

```java
private static Optional<Response> getChannelResponse(ChannelHandlerContext ctx) {
     return Optional.ofNullable(ctx.channel().attr(CHANNEL_RESPONSE_KEY).get());
}

// option 1 StreamingResponseData constructor needs to deal with Optional
Response response = new StreamingResponseData(getChannelResponse(ctx), new Http2StreamingData(headers));

// option 2 wrapResponse needs to deal with Optional
Response response = new StreamingResponseData(
  getChannelResponse(ctx).orElseThrow(() -> return new RuntimeException("Couldn't find Response for ctx: " + ctx)), 
  new Http2StreamingData(headers)
);
```

src/main/java/com/xjeffrose/xio/http/Http2ClientCodec.java
pdex: Throwing in the pipeline pretty much means we're fucked and the connection is going to be closed. I'm okay with that for situations like this where we've reached a bad state and things just need to die.

PR #167, opened 2018-01-07
reverse proxy functional test

pdex:


src/main/java/com/xjeffrose/xio/http/Headers.java
```diff
@@ -4,17 +4,27 @@
...
+import io.netty.handler.codec.http2.DefaultHttp2Headers;
+import io.netty.handler.codec.http2.Http2Headers;
 
 @UnstableApi
 public interface Headers
     extends io.netty.handler.codec.Headers<CharSequence, CharSequence, Headers>,
         Iterable<Entry<CharSequence, CharSequence>> {
 
-  default HttpHeaders http1Headers() {
+  default HttpHeaders http1Headers(boolean isTrailer, boolean isRequest) {
```
jkinkead: Unused parameters? Maybe document?

src/main/java/com/xjeffrose/xio/http/ProxyBackendHandler.java
```diff
@@ -0,0 +1,70 @@
...
+    if (ctx.pipeline().get(Http2Handler.class) != null) {
+      log.debug("handlerAdded: adding Http2StreamMapper");
+      // we are an http2 pipeline
+      ctx.pipeline().addBefore("application codec", "stream mapper", new Http2StreamMapper());
+    }
+  }
+
+  @Override
+  public void channelRead(ChannelHandlerContext ctx, Object msg) throws Exception {
+    log.debug("RawBackendHandler[{}] channelRead: {}", this, msg);
```
jkinkead: Hey look @xjdr , this seems to be how you do log formatting.

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
```diff
@@ -0,0 +1,92 @@
...
+    public ChannelFuture apply(BiFunction<Object, ChannelPromise, ChannelFuture> write) {
+      return write.apply(object, promise);
+    }
+  }
+
+  List<ObjectAndPromise> writeBuffer = new ArrayList<>();
+
+  @Override
+  public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+    if (!(evt instanceof WriteReady)) {
```
jkinkead: Why not `evt == WriteReady.INSTANCE`?

Also, why not just:
```java
public static final Object WRITE_READY = new Object();
```
?

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
```diff
@@ -0,0 +1,92 @@
...
+      this.object = object;
+      this.promise = promise;
+    }
+
+    public ChannelFuture apply(BiFunction<Object, ChannelPromise, ChannelFuture> write) {
+      return write.apply(object, promise);
+    }
+  }
+
+  List<ObjectAndPromise> writeBuffer = new ArrayList<>();
```
jkinkead: Any reason for this not to be `private final`?

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
```diff
@@ -0,0 +1,92 @@
...
+  @Override
+  public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+    if (!(evt instanceof WriteReady)) {
+      ctx.fireUserEventTriggered(evt);
+      return;
+    }
+
+    // drain buffer
+    while (writeBuffer.size() > 0) {
+      ObjectAndPromise pair = writeBuffer.remove(0);
```
jkinkead: `O(n)` for `ArrayList`. Use a `LinkedList` instead?

src/main/java/com/xjeffrose/xio/http/internal/FullHttp2Response.java
```diff
@@ -23,9 +25,9 @@ public FullHttp2Response(Http2Headers delegate, int streamId) {
   @Override
   public HttpResponseStatus status() {
     try {
-      return HttpResponseStatus.valueOf(Integer.parseInt(delegate.status().toString()));
-    } catch (Exception e) {
-      return null;
+      return HttpConversionUtil.parseStatus(delegate.status());
+    } catch (Http2Exception e) {
+      throw new RuntimeException(e);
```
jkinkead: Document exception thrown?

src/main/java/com/xjeffrose/xio/http/internal/Http2HeadersWrapper.java
```diff
@@ -578,10 +582,15 @@ public int size() {
...
-  public HttpHeaders http1Headers() {
-    return null;
+  public HttpHeaders http1Headers(boolean isTrailer, boolean isRequest) {
+    try {
+      HttpHeaders headers = new DefaultHttpHeaders();
+      HttpConversionUtil.addHttp2ToHttpHeaders(
+          -1, delegate, headers, HttpVersion.HTTP_1_1, isTrailer, isRequest);
+      return headers;
+    } catch (Http2Exception e) {
+      throw new RuntimeException(e);
```
jkinkead: Document, maybe by adding to the interface documentation?

This should probably be a custom exception, also...

src/test/java/com/xjeffrose/xio/fixtures/JulBridge.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.List;
+import com.xjeffrose.xio.client.ClientConfig;
+import org.slf4j.bridge.SLF4JBridgeHandler;
+import org.junit.BeforeClass;
+
+public abstract class JulBridge {
+  private static int dummy = -1;
+
+  public static void initialize() {
+    if (dummy == -1) {
```
jkinkead: Not thread-safe; probably whole method should be synchronized, else you could initialize this twice, which I assume is bad.

I'm also not sure the JVM guarantees that `dummy` will be nonzero when this method is entered for the first time, especially if multiple threads hit it at once.

src/test/java/com/xjeffrose/xio/fixtures/OkHttpUnsafe.java
```diff
@@ -16,6 +16,8 @@
 import javax.net.ssl.X509TrustManager;
 import okhttp3.OkHttpClient;
 import okhttp3.mockwebserver.MockWebServer;
+import java.util.Arrays;
```
jkinkead: Sort?

src/test/java/com/xjeffrose/xio/fixtures/JulBridge.java
jkinkead: The JLS [does guarantee initialization of nonfinal static fields](https://docs.oracle.com/javase/specs/jls/se8/html/jls-12.html#jls-12.4.2), FWIW, even if viewed from a thread not charged with loading the class.

src/main/java/com/xjeffrose/xio/application/ApplicationConfig.java
```diff
@@ -12,8 +12,7 @@
 @Slf4j
 public class ApplicationConfig {
 
-  private final Config config;
-
+  @Getter private final Config config;
```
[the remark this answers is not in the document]
pdex: good catch, this was a hack I made to get to something from a test. I'll put this on the chopping block.

src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -0,0 +1,104 @@
...
+import com.xjeffrose.xio.core.XioIdleDisconnectHandler;
+import java.util.function.Supplier;
+import java.util.concurrent.Future;
+import io.netty.channel.ChannelFutureListener;
+
+public class Client {
+
+  private final ClientState state;
+  private final Supplier<ChannelHandler> appHandler;
+  Channel channel;
```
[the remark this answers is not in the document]
pdex: 1) This should be private.

2) I would prefer for this to be final as well but the netty idiom with the `Bootstrap` is to both create the `Channel` and start it doing something. So the two options that I see are:
 a) Have the constructor do more than construct so we can set `final Channel channel`
 b) Defer the construction of `Channel channel` knowing that we will probably never update it again after setting it once.

I'm going with option b) for the time being because it should make testing easier. It may be possible to create a compromise between a) and b) but that would require digging into the Bootstrap class and moving some of it's innards into the `Client` class.

src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -0,0 +1,104 @@
...
+                .addLast(
+                    "negotiation handler",
+                    new HttpClientNegotiationHandler(Client.this::buildHttp2Handler))
+                .addLast("codec", CodecPlaceholderHandler.INSTANCE)
+                .addLast("application codec", ApplicationCodecPlaceholderHandler.INSTANCE)
+                .addLast("idle handler", new XioIdleDisconnectHandler(60, 60, 60))
+                .addLast("message logging", new XioMessageLogger(Client.class, "objects"))
+                .addLast("request buffer", new RequestBuffer())
+                .addLast("app handler", appHandler.get());
+            //              .addLast(new RawBackendHandler(ctx));
```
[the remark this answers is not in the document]
pdex: Yup in the kill ring.

src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -0,0 +1,104 @@
...
+              // log?
+            } else {
+              // log?
+              // fail the write future?
+              throw new RuntimeException(f.cause());
+            }
+          };
+      channel = future.channel();
+      // connect().addListener(f -> connected((ChannelFuture) f));
+      connect().addListener(l);
```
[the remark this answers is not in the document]
pdex: Yup copy/paste bug. Fixed.

src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -0,0 +1,104 @@
...
+          f -> {
+            if (f.isDone() && f.isSuccess()) {
+              // log?
+            } else {
+              // log?
+              // fail the write future?
+              throw new RuntimeException(f.cause());
+            }
+          };
+      channel = future.channel();
```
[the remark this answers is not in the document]
pdex: See below for thread commentary.

src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -0,0 +1,104 @@
...
+      // connect().addListener(f -> connected((ChannelFuture) f));
+      connect().addListener(l);
+    }
+    return channel.writeAndFlush(request);
```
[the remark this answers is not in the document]
pdex: This is a goofy netty thing.

> Returns a channel where the I/O operation associated with this future takes place.

In my experience this means you can get the `Channel` from the `ChannelFuture` as soon as it's `connect()`ing or `bind()`ing or what have you. That being said, this works and is thread safe as long as the thread on the `EventLoopGroup` that this channel is associated with is the same thread that's invoking `write`. I'm guaranteeing that right now. In the future we need a more complex setup that chains the futures like so:
```
connect().andThen(write).andThen(write)
```

Netty `ChannelFutureListener` doesn't support this syntax, but you can do something close to it. We'll also want to bundle the initial connect and write promises together since the method only returns a single `ChannelFuture`

src/main/java/com/xjeffrose/xio/http/DefaultFullRequest.java
```diff
@@ -59,6 +68,6 @@ public Builder host(String host) {
   }
 
   static Builder builder() {
-    return new AutoValue_DefaultFullRequest.Builder();
+    return new AutoValue_DefaultFullRequest.Builder().streamId(-1);
```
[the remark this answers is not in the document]
pdex: Good catch. It's on my TODO list to hide this behind a constant.

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
pdex: Fixed.

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
pdex: Following netty convention here:
https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2ConnectionPrefaceAndSettingsFrameWrittenEvent.java

Although they seem to be conflicted on equals versus instanceof 
https://github.com/netty/netty/blob/4.1/codec-http2/src/main/java/io/netty/handler/codec/http2/Http2FrameCodec.java#L233-L238

With this:
```
  public static class WriteReady {
    public static final WriteReady INSTANCE = new WriteReady();
  }
```
If you have a message logger in your pipeline it will log that an event was fired of type `WriteReady`. You lose that information with `Object WRITE_READY`. I should probably make the constructor private.

There's probably a more in depth conversation to be had about the speed of `==` vs `instanceof`. I have no preference here nor the desire to micro benchmark at this point in time. :)

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
pdex: https://docs.oracle.com/javase/7/docs/api/java/util/ArrayList.html

> The size, isEmpty, get, set, iterator, and listIterator operations run in constant time. The add operation runs in amortized constant time, that is, adding n elements requires O(n) time. All of the other operations run in linear time (roughly speaking). The constant factor is low compared to that for the LinkedList implementation.

Are you thinking of [ArrayDeque](https://docs.oracle.com/javase/7/docs/api/java/util/ArrayDeque.html) ?

src/main/java/com/xjeffrose/xio/http/internal/Http2HeadersWrapper.java
pdex: Documentation, yes. Custom Exception? I dunno. I'm just trying to quash the checked exception here. We should maybe establish a policy around checked exceptions.

src/test/java/com/xjeffrose/xio/fixtures/JulBridge.java
pdex: I'll make this `final`, and the method `synchronized`. I'm not super concerned if this runs twice as I'm just trying to shutup the `java.util.logging` messages that `MockWebServer` emits.

src/test/java/com/xjeffrose/xio/fixtures/OkHttpUnsafe.java
pdex: I'm surprised the pre-commit git-hook missed that.

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
jkinkead: `remove(0)` is linear-time per the docs you quote. Suggestion still stands?

src/main/java/com/xjeffrose/xio/http/internal/Http2HeadersWrapper.java
jkinkead: Checked exceptions kind of suck, but they're deeply ingrained in Java libraries and general standards. It's accepted style, and we probably shouldn't fight that without good reason.

Regardless of whether it's checked or not, it should probably be a custom type.

src/main/java/com/xjeffrose/xio/http/Headers.java
pdex: comments added

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
pdex: Fair enough argument, I'll change to eq.

src/main/java/com/xjeffrose/xio/http/RequestBuffer.java
pdex: I meant should we use the ArrayDeque, not LinkedList.

PR #175, opened 2018-02-20
Broke out testing utility classes into xio-test

pdex:


xio-test/src/main/java/com/xjeffrose/xio/test/OkHttpUnsafe.java
```diff
@@ -0,0 +1,112 @@
...
+import javax.net.ssl.SSLContext;
+import javax.net.ssl.SSLSession;
+import javax.net.ssl.SSLSocketFactory;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.X509TrustManager;
+import okhttp3.OkHttpClient;
+import okhttp3.Protocol;
+import okhttp3.mockwebserver.MockWebServer;
+
+public class OkHttpUnsafe {
```
jkinkead: I feel like this and the public accessors need to have very scary docs telling people never to use them outside of testing. . .

PR #179, opened 2018-03-14
Updated RoutePartial to use RouteState

pdex:


xio-core/src/main/java/com/xjeffrose/xio/http/RoutePartial.java
```diff
@@ -1,22 +1,26 @@
...
+  RoutePartial(Request request, String path, RouteState route) {
     this.request = request;
+    this.path = path;
     this.route = route;
-    this.handler = handler;
   }
 
   void apply(ChannelHandlerContext ctx) {
-    handler.handle(ctx, request, route);
+    route.handler().handle(ctx, request, route);
```
jguamie: Is handler() guaranteed to not be null here? Does lombok take care of this for us?

xio-core/src/main/java/com/xjeffrose/xio/http/RouteState.java
```diff
@@ -25,4 +26,17 @@ public RouteState(
   public RouteState(RouteConfig config, PipelineRequestHandler handler) {
     this(RouteState::buildRoute, config, handler);
   }
+
+  public static RouteState defaultRoute(PipelineRequestHandler handler) {
+    RouteConfig config = new RouteConfig(ConfigFactory.load().getConfig("xio.defaultRoute"));
+    return new RouteState((c) -> Route.build("*"), config, handler);
```
jguamie: c is not used here. Do we replace this with "ignored"?

xio-core/src/main/java/com/xjeffrose/xio/http/RouteState.java
```diff
@@ -25,4 +26,17 @@ public RouteState(
   public RouteState(RouteConfig config, PipelineRequestHandler handler) {
     this(RouteState::buildRoute, config, handler);
   }
+
+  public static RouteState defaultRoute(PipelineRequestHandler handler) {
+    RouteConfig config = new RouteConfig(ConfigFactory.load().getConfig("xio.defaultRoute"));
```
jguamie: Should all these string literal put into constants somewhere? e.g., "xio.defaultRoute" and "*" on the next line.

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
```diff
@@ -103,29 +89,48 @@ public RouteStates(Function<U, T> factory, RouteConfigs configs) {
...
+    private final RouteStates<ProxyRouteState> routeStates;
     private final ProxyClientFactory clientFactory;
 
+    public <T, K, U> Collector<T, ?, Map<K, U>> toLinkedMap(
+        Function<? super T, ? extends K> keyMapper, Function<? super T, ? extends U> valueMapper) {
+
+      return Collectors.toMap(
+          keyMapper,
+          valueMapper,
+          (u, v) -> {
```
jguamie: What is the 'v' in this context? Could we add comments on what this is? Does the 'u' here map to the generic 'U' in the method declaration?

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
```diff
@@ -103,29 +89,48 @@ public RouteStates(Function<U, T> factory, RouteConfigs configs) {
...
-              (ProxyRouteConfig c) ->
-                  new ProxyRouteState(this, c, new ProxyHandler(clientFactory, c)),
-              config.routeConfigs);
+          new RouteStates<ProxyRouteState>(
+              ImmutableMap.copyOf(
+                  config
+                      .routeConfigs
+                      .stream()
+                      .map(
+                          (ProxyRouteConfig c) ->
```
jguamie: Rename 'c' to "config"?

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
```diff
@@ -103,29 +89,48 @@ public RouteStates(Function<U, T> factory, RouteConfigs configs) {
...
+                              new ProxyRouteState(this, c, new ProxyHandler(clientFactory, c)))
+                      .collect(toLinkedMap(s -> s.path(), s -> s))));
```
jguamie: 's' could be more descriptive as well for a variable name.

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
```diff
@@ -103,29 +89,48 @@ public RouteStates(Function<U, T> factory, RouteConfigs configs) {
...
+
     public EdgeProxyState(EdgeProxyConfig config) {
       super(config);
       clientFactory = new ProxyClientFactory(this);
       routeStates =
-          new RouteStates<ProxyRouteState, ProxyRouteConfig>(
-              (ProxyRouteConfig c) ->
-                  new ProxyRouteState(this, c, new ProxyHandler(clientFactory, c)),
-              config.routeConfigs);
+          new RouteStates<ProxyRouteState>(
```
jguamie: There is a lot of chaining happening here. Can we have a comment describing what you're accomplishing here?

xio-core/src/test/java/com/xjeffrose/xio/http/GrpcFunctionalTest.java
```diff
@@ -375,15 +379,22 @@ public Client createClient(ChannelHandlerContext ctx, ClientConfig config) {
...
           }
         };
+
+    ApplicationState appState =
+        new ApplicationState(ApplicationConfig.fromConfig("xio.defaultApplication"));
+    XioServerConfig serverConfig = XioServerConfig.fromConfig("xio.testGrpcServer");
+
     XioServerBootstrap bootstrap =
-        XioServerBootstrap.fromConfig("xio.testGrpcServer")
+        new XioServerBootstrap(appState, serverConfig, new XioServerState(serverConfig))
```
jguamie: White spacing issues. Also, a comment to describe what you're doing here would help. Lots of chain magic happening here.

jguamie:
It would be nice if we had contexts/section descriptions broken out within each of the test classes so it is easier to read what the tests are verifying what and where.

jguamie:
Also, I saw we do a lot of method chaining throughout. Law of Demeter recommends we don't do this.

xio-core/src/main/java/com/xjeffrose/xio/http/PipelineRouter.java
```diff
@@ -10,11 +10,11 @@
   private final PathToRequestHandler requestHandlers;
 
   public PipelineRouter(
-      ImmutableMap<Route, PipelineRequestHandler> routes, PipelineRequestHandler defaultHandler) {
+      ImmutableMap<String, RouteState> routes, PipelineRequestHandler defaultHandler) {
```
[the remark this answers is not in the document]
pdex: We do! We're using the google style formatter: https://github.com/google/google-java-format

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
```diff
@@ -78,7 +79,11 @@ public Request buildRequest(Request request, String proxyHost, String path) {
     return result;
   }
 
-  public void handle(ChannelHandlerContext ctx, Request request, Route route) {
+  private void appendXForwardedFor(Request request) {
```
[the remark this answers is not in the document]
pdex: it's this: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
```diff
@@ -78,7 +79,11 @@ public Request buildRequest(Request request, String proxyHost, String path) {
...
+    // TODO(CK): update request headers
+  }
+
+  public void handle(ChannelHandlerContext ctx, Request request, RouteState route) {
```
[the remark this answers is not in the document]
pdex: I'm 99% certain that this is no longer WIP. Between `Request` and `RouteState` we should be able to attach any information that we need about the request/route that need to be handled.

I did however neglect to `@Override` the method.

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
```diff
@@ -34,9 +34,10 @@ public String buildProxyHost(Request request, ClientConfig clientConfig) {
     }
   }
 
-  public String buildProxyPath(Request request, Route route) {
+  public String buildProxyPath(Request request, RouteState state) {
     Optional<String> pathSuffix =
-        route
+        state
```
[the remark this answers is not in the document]
pdex: are you suggesting `Optional<String> pathSuffix = state` instead?

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
```diff
@@ -102,6 +107,8 @@ public void handle(ChannelHandlerContext ctx, Request request, Route route) {
     String proxyHost = buildProxyHost(request, clientConfig);
     Request proxyRequest = buildRequest(request, proxyHost, buildProxyPath(request, route));
 
+    appendXForwardedFor(proxyRequest);
+
     factory.getClient(ctx, clientConfig).write(proxyRequest);
```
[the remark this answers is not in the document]
pdex: good call. How's Demeter these days?

xio-core/src/main/java/com/xjeffrose/xio/http/RoutePartial.java
pdex: No, this is a java feature: https://github.com/xjdr/xio/blob/9391ca706467d3a88c71e8a3dd557dd4a140bd0d/xio-core/src/main/java/com/xjeffrose/xio/http/RouteState.java#L13

https://en.wikipedia.org/wiki/Final_(Java)#Final_variables

Technically the caller to `new RouteState` could pass in null, but that would be programmer error. We could put some precondition checks into the RouteState constructor but I generally like to avoid that as it makes unit tests very verbose when you're not allowed to pass in a null value. 

The flipside to that is using a builder pattern in your unit tests so that the state object is constructed correctly.

I could go either way here.

xio-core/src/main/java/com/xjeffrose/xio/http/RouteState.java
pdex: my rule of thumb with constants is to create one whenever I'm about to copy/paste a value.

xio-core/src/main/java/com/xjeffrose/xio/http/RouteState.java
pdex: good call.

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
pdex: https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toMap-java.util.function.Function-java.util.function.Function-java.util.function.BinaryOperator-java.util.function.Supplier-

This is the merge function that gets called by `toMap` when the stream produces two entries with the same key. For us this is an error condition. But there's a conversation to be had about how this needs to go away in favor of a "route map" that doesn't care about the ordering of the keys.

To your previous point, `v` could be `ignored`.

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
pdex: yup.

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
pdex: Yeah, I should probably get into the habit of commenting calls to `stream()`

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
pdex: should be `state`

xio-core/src/test/java/com/xjeffrose/xio/http/GrpcFunctionalTest.java
pdex: We should document a rule of thumb about number of chains allowed before comments should be added.

xio-core/src/test/java/com/xjeffrose/xio/http/GrpcFunctionalTest.java
```diff
@@ -6,10 +6,14 @@
 import com.typesafe.config.ConfigFactory;
```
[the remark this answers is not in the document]
pdex: "variation?"

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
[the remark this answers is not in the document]
jguamie: Yes, exactly. And have .route() be the next line and such.

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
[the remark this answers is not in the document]
jguamie: Ok, I would suggest cleaning up the comments in this handle() method then. There's the comment block at the top that doesn't have an explanation on why it is commented out.

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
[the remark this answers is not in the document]
jguamie: Understood. Thanks!

xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
jguamie: Ok, great. Let's do 'ignored' then.

xio-core/src/test/java/com/xjeffrose/xio/http/GrpcFunctionalTest.java
jguamie: Agreed.

PR #185, opened 2018-03-21
wk/TACO-11 WIP

khappucino:
WIP for initial feedback


pdex:
So here's how tracing works in my mind:
An Application could have either Clients or Servers or both. There should only really be one XioTracing per Application.

XioTracing acts as an object factory with state about the location of zipkin. It has `newServerHandler` and `newClientHandler` those can be used by the Application to setup handlers on a server or client with zipkin tracing.

PR #190, opened 2018-03-28
Dc/taco 11 test update

khappucino:
based on failing test during CI and after some discussion I am nuking the ClientTest and extracting the ChannelInitializer work into a ClientChannelInitializer class that is responsible for just building the pipeline. This is tested by adding the subject's output to an EmbeddedChannel then asking the channel if it has the expeced traceHandler


xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelInitializer.java
```diff
@@ -0,0 +1,70 @@
...
+public class ClientChannelInitializer {
+  public static final String sslHandlerName = "ssl handler";
+  public static final String negotiationHandlerName = "negotiation handler";
+  public static final String codecName = "codec";
+  public static final String tracingName = "distributed tracing";
+  public static final String applicationCodecName = "application codec";
+  public static final String idleHandlerName = "idle handler";
+  public static final String messageLoggingName = "message logging";
+  public static final String requestBufferName = "request buffer";
+  public static final String appHandlerName = "app handler";
```
pdex: is this an intelliJ recommendation? looks like overkill to me, we're not manipulating the pipeline using these labels.

xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelInitializer.java
```diff
@@ -0,0 +1,70 @@
...
+  private final XioTracing tracing;
+
+  public ClientChannelInitializer(
+      ClientState state, Supplier<ChannelHandler> appHandler, XioTracing tracing) {
+    this.state = state;
+    this.appHandler = appHandler;
+    this.tracing = tracing;
+  }
+
+  public ChannelInitializer createChannelInitializer() {
```
pdex: unless it's making testing easier, I'm not seeing a good reason for composing with `ChannelInitializer` instead of inheiriting it. Right now `ClientChannelInitializer` is a `ChannelInitializer` Factory and not a `ChannelInitializer`.

xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelInitializer.java
khappucino: It was just for the unit testing but I can unwind that since it's mainly extra indirection

xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelInitializer.java
khappucino: ok I can buy that, adjusting

PR #192, opened 2018-03-29
TACO-35 - adding span to Request object

manimaul:
adding a `Span` property to the Request interface for trace propagation


xio-core/src/main/java/com/xjeffrose/xio/http/StreamingRequestData.java
```diff
@@ -62,6 +72,20 @@ public ByteBuf body() {
     return request.body();
   }
 
+  // endregion
```
manimaul: i'm not sure if this (pragma separating multiple interface implementations) is consistent with the project style

PR #195, opened 2018-04-02
dc/clientconnectbeforewrite serializing the connect before calling wr…

khappucino:
dc/clientconnectbeforewrite serializing the connect before calling write, I noticed that if you call write concurrently with connect, there is a race condition where we grab the channelPipeline's tail reference and we have not added the channel handlers from clientChannelInitializer yet (initChannel gets called after we try to write sometimes)

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
@@ -7,9 +7,6 @@ import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelPromise;
-import io.netty.util.concurrent.Future;
-import io.netty.util.concurrent.GenericFutureListener;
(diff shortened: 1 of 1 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -56,13 +59,33 @@ public ChannelFuture write(Request request) {
       ChannelFuture future = connect();
       channel = future.channel();
       ChannelPromise promise = channel.newPromise();
-      PromiseCombiner combiner = new PromiseCombiner();
-      combiner.add(future.addListener(connectionListener));
-      combiner.add(channel.writeAndFlush(request).addListener(writeListener));
-      combiner.finish(promise);
+      future.addListeners(connectionListener, new GenericFutureListener<Future<? super Void>>() {
```
manimaul: a lambda would be preferred rather than an anonymous GenericFutureListener

manimaul: too bad `ChannelFuture` does not have a monadic interface - a flatMap or andThen would make this easier to read

PR #196, opened 2018-04-03
TACO-35 zipkin duplex handler

manimaul:
Adding a client duplex handler which:
sends zipkin trace on  outbound `Request`  when  `startOfStream`
sends zipkin trace on  inbound `Response` when `endOfStream` 

Adding a server duplex handler which:
sends zipkin trace on inbound `Request` interface  when  `startOfStream`
sends zipkin trace on outbound `Response` interface when `endOfStream`


xio-core/src/main/java/com/xjeffrose/xio/http/DefaultFullResponse.java
```diff
@@ -39,16 +41,19 @@ public String version() {
...
 
-    abstract Optional<TraceInfo> httpTraceInfo();
-
-    abstract DefaultFullResponse autoBuild();
-
     public DefaultFullResponse build() {
       if (!httpTraceInfo().isPresent()) {
         httpTraceInfo(new TraceInfo());
       }
+      endOfStream(true);
```
pdex: Yeah this is probably better for readability than checking object instance.

xio-core/src/main/java/com/xjeffrose/xio/http/Headers.java
```diff
@@ -6,12 +6,26 @@
...
+  @Nullable
+  default String get(String name) {
+    CharSequence value = get((CharSequence) name);
+    if (value != null) {
+      return value.toString();
+    } else {
+      return null;
+    }
+  }
+
```
pdex: nice use of default

xio-core/src/main/java/com/xjeffrose/xio/tracing/HttpTracingState.java
```diff
@@ -1,38 +1,19 @@
...
-    ctx.channel().attr(context_key).set(context);
-  }
-
-  // get the context set by outbound request XioRequestEncoder (outbound)
-  public static TraceContext popContext(ChannelHandlerContext ctx) {
-    return ctx.channel().attr(context_key).getAndSet(null);
+  public static Optional<Span> popSpan(ChannelHandlerContext ctx) {
+    return Optional.ofNullable(ctx.channel().attr(span_key).getAndSet(null));
   }
 }
```
pdex: now that we have the tracing information in the request/response objects we should consider removing this concept of attaching it to the `ctx`.

manimaul: yeah - i'm just not sure how we would handle `HttpClientResponseTracingHandler.exceptionCaught` since we do not have a reference to Response

PR #198, opened 2018-04-05
Defer assignment of the updater

pdex:


xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
```diff
@@ -99,8 +100,16 @@ public T init(String file) {
     return metadata.value;
   }
 
-  public void start() {
+  @VisibleForTesting
+  public void setUpdater(BiConsumer<T, T> updater) {
```
manimaul: maybe make this package private or protected

xio-core/src/test/java/com/xjeffrose/xio/config/ConfigReloaderUnitTest.java
```diff
@@ -39,22 +39,53 @@ String createConfig(String value) throws IOException, FileNotFoundException {
...
+          @Override
+          public void fireUpdated() {
+            assertEquals(20, this.config.limit);
+            executor.shutdown();
+          }
+        };
+    assertEquals(10, config.limit);
+    reloader.start(state::update);
+
+    Thread.sleep(2000);
```
manimaul: CountDownLatch might be better than a sleep

pdex: I just tried to replace the sleeps with countdown latches, and that's when I was brutally reminded that we're checking the modification date of a file in the implementation. To do this right without sleeps we'll need to mock out that modification check. Not super concerned about this at the moment we can throw this into a tech debt backlog.

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
pdex: done

PR #200, opened 2018-04-06
dc/TACO-12 Updating the HttpClientTracingHandlerIntegrationTest for great justice

khappucino:
Updating the HttpClientTracingHandlerIntegrationTest to use Client.  I had to modify XioTracing to expose the method buildReporter so that we can replace the buildReporter during tests. This is the method that I found used in Brave integration tests (overwriting the reporter and recording the spans). I also fixed a bug that I introduced when I added the tracinghandlers, the ApplicationCodec is now right after the HttpCodec, I accidentally put the tracing handler between the HttpCode and the ApplicationCodec.  I updated the application.conf to have a sample rate of 100% since the existing 50% sampling rate makes the tests fail half the time. LOL. The HttpClientTracingDispatch was modified to use the span that was shared via the channelcontextAttribute (setSpan) and we still add the span to the response. I modified the Http1Response class to change the endOfStream method to return false. This is due to the fact that netty HttpCodec always generates a LastHttpContent message which is actually the end of stream.

The Client write method has been modified because if you create a client and invoke write, there is a race condition between the connection (initChannel) and the write command. What we see is that the first write you try will be resolved to the ChannelPipeline's HEAD context, since the channel handlers have not been added yet. This doesn't seem to be a problem in the EdgeProxy configuration, however I did notice that in the EdgeProxyTest, the Client handle method (and thus the write method) gets called TWICE in quick succession. I don't think this is on purpose is it?  It looks like the first call doesn't make it but the second call actually hits the channel handlers since they have been added by now.  The client write looks gross now but I don't think there is another way to ensure that the connection and channel handlers are ready before writing. If you try to write twice back to back you may send before a) the connection is made b) the channel handlers are initialized.


xio-core/src/main/java/com/xjeffrose/xio/http/internal/Http1Response.java
```diff
@@ -33,7 +33,7 @@ public Http1Response(HttpResponse delegate) {
 
   @Override
   public boolean endOfStream() {
-    return true;
+    return false;
```
manimaul: these xio message classes could use some documentation and/or better names

xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -53,16 +55,47 @@ private ChannelFuture connect() {
...
-      ChannelFuture future = connect();
-      channel = future.channel();
+      connectionFuture = connect();
+      channel = connectionFuture.channel();
       ChannelPromise promise = channel.newPromise();
-      PromiseCombiner combiner = new PromiseCombiner();
-      combiner.add(future.addListener(connectionListener));
-      combiner.add(channel.writeAndFlush(request).addListener(writeListener));
-      combiner.finish(promise);
+      connectionFuture.addListeners(
```
manimaul: a comment or two here might be useful

xio-core/src/main/java/com/xjeffrose/xio/http/internal/Http1Response.java
pdex: I think we should make some sample/example code to help us:
1) refine the look and feel of the interfaces
2) point at from a user guide
3) we should actually add javadoc too :)

xio-core/src/test/java/com/xjeffrose/xio/tracing/HttpClientTracingHandlerIntegrationTest.java
```diff
@@ -0,0 +1,180 @@
...
+  Logger hush = disableJavaLogging();
+  CompletableFuture<Response> local = new CompletableFuture<Response>();
+  MockWebServer server;
+
+  private class FakeTracer extends XioTracing {
+    public FakeTracer(Config config) {
+      super(config);
+    }
+
+    @Override
```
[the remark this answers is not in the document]
pdex: I don't mind doing this, I try to use Mocks only when it's absolutely necessary.

xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelResponse.java
```diff
@@ -0,0 +1,20 @@
...
+
+  @Getter private boolean isSuccess;
+  @Getter private Throwable cause;
+
+  public ClientChannelResponse(boolean isDone, boolean isSuccess, @Nullable Throwable cause) {
+    this.isDone = isDone;
+    this.isSuccess = isSuccess;
+    this.cause = cause;
+  }
+}
```
pdex: If this only exists while the class is under test, we can make it an inner class of the test itself. If we need it in another class we can put it in `src/test/java/com/xjeffrose/xio/http`

xio-core/src/test/java/com/xjeffrose/xio/tracing/HttpClientTracingHandlerIntegrationTest.java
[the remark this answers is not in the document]
khappucino: kewl

xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelResponse.java
khappucino: ok i can do that =p

xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -65,4 +69,38 @@ public ChannelFuture write(Request request) {
...
+    internalConnect()
+        .addListeners(
+            connectionListener,
+            (resultFuture) -> {
+              val response =
+                  new ClientChannelResponse(
+                      resultFuture.isDone(), resultFuture.isSuccess(), resultFuture.cause());
+              outerResult.complete(response);
+            });
+    return outerResult;
```
pdex: Why can't we return the ChannelFuture from `internalConnect`?

khappucino: actually in this case we can definitly return the channel future, the issue i think was more on the write(Request request) because if the channel is null you can't new ChannelPromise() and resolve it with a throwable cause

khappucino: I can change the connect() method to return the ChannelFuture

xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelResponse.java
khappucino: ok after discussing with pdex, this can stay

PR #202, opened 2018-04-06
Wk/taco 35 zipkin server integration test

manimaul:
added test to verify span is sent


xio-core/src/test/java/com/xjeffrose/xio/fixtures/OkHttpUnsafe.java
```diff
@@ -32,15 +32,6 @@
     return keyManagerFactory.getKeyManagers();
   }
 
-  public static KeyManager[] getEmptyKeyManager() throws Exception {
```
manimaul: this function is unused and a duplicate

pdex: This whole class can be removed and the one in xio-test can be used instead.

xio-core/src/main/java/com/xjeffrose/xio/http/StreamingRequestData.java
```diff
@@ -16,7 +16,7 @@
   public StreamingRequestData(Request request, StreamingData data, TraceInfo traceInfo) {
     this.request = request;
     this.data = data;
-    this.traceInfo = traceInfo;
+    this.traceInfo = traceInfo == null ? new TraceInfo(request.headers()) : traceInfo;
```
pdex: 👍 nice ternary operator

xio-core/src/main/java/com/xjeffrose/xio/tracing/XioTracing.java
```diff
@@ -14,13 +14,13 @@
 
 public class XioTracing {
 
-  private final Tracing tracing;
+  protected final Tracing tracing;
```
pdex: Can we replace this with a @VisibleForTesting getter?

PR #203, opened 2018-04-09
fix jdk9 build

pdex:


xio-core/pom.xml
```diff
@@ -243,22 +250,29 @@
...
+      <artifactId>powermock-core</artifactId>
+      <version>2.0.0-beta.5</version>
       <scope>test</scope>
     </dependency>
 
    <dependency>
       <groupId>org.powermock</groupId>
-      <artifactId>powermock-api-mockito</artifactId>
-      <version>1.6.5</version>
+      <artifactId>powermock-api-mockito2</artifactId>
```
manimaul: i wonder if we still need powermock - https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable

README.md
```diff
@@ -49,3 +49,17 @@ following maven plugins maintain the source code standards:
...
+
+### Testing new jdk versions
+
+#### build the docker image
+
+docker build -t xio-jdk9 .
+
+#### test the build
+
+docker run  --mount type=bind,src=`pwd`,dst=/home -t xio-jdk9 ./mvnw verify
```
manimaul: i think the recommended way to do this now is with --volume rather than --mount

pdex: /shrug docker

xio-core/pom.xml
pdex: I'm fine with that, we should probably have some guidelines around when/how we mock things.

PR #205, opened 2018-04-10
disable packagecloud in circleci config

manimaul:
title says it

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -23,4 +23,4 @@ deployment:
     branch: master
     commands:
       - mvn -N io.takari:maven:wrapper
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -23,4 +23,4 @@ deployment:
     branch: master
     commands:
       - mvn -N io.takari:maven:wrapper
-      - ./mvnw -pl xio-core -P packagecloud -s .circleci.settings.xml -DskipTests deploy
+#      - ./mvnw -pl xio-core -P packagecloud -s .circleci.settings.xml -DskipTests deploy
```
pdex: we can just remove all of these commands.

PR #206, opened 2018-04-10
Upgrade to circle ci 2.0

pdex:


xio-core/pom.xml
```diff
@@ -464,6 +464,13 @@
         <groupId>org.eluder.coveralls</groupId>
         <artifactId>coveralls-maven-plugin</artifactId>
         <version>4.3.0</version>
+        <dependencies>
+          <dependency>
+            <groupId>javax.xml.bind</groupId>
```
manimaul: is this for java9?

pdex: yes they took this out of the JVM, it's how a lot of projects were doing base64 without pulling in another jar.

PR #207, opened 2018-04-11
Wk/taco 131 just some names

manimaul:
Since we are representing h1 and h2 messages through a single interface - we should be clear about the following differences:
* connection stream id
* message stream id
In addition, it should be clear if the `Message` is the start or end of the message stream.

Definitions: 

> Stream: A bidirectional flow of bytes within an established connection, which may carry one or more messages.
> Message: A complete sequence of frames that map to a logical request or response message.
> Frame: The smallest unit of communication in HTTP/2, each containing a frame header, which at a minimum identifies the stream to which the frame belongs.

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
@@ -1,0 +1,14 @@
+package com.xjeffrose.xio.http;
+
+/**
+ * Interface representing a HTTP1/2 Message (Request/Response) Message: A complete sequence of
+ * frames that map to a logical request or response message. See: <a href= ></a>
+ */
+public interface Message {
+
+  int STREAM_ID_NONE = -1;
+
+  boolean startOfMessage();
+
+  boolean endOfMessage();
+}
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
@@ -5,7 +5,8 @@ import javax.annotation.Nullable;
 
 /**
  * Interface representing a HTTP1/2 Message (Request/Response) Message: A complete sequence of
- * frames that map to a logical request or response message. See: <a href= ></a>
+ * frames that map to a logical request or response message.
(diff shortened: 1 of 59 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
```diff
@@ -0,0 +1,14 @@
+package com.xjeffrose.xio.http;
+
+/**
+ * Interface representing a HTTP1/2 Message (Request/Response) Message: A complete sequence of
+ * frames that map to a logical request or response message. See: <a href= ></a>
```
pdex: link to https://xjdr.github.io/xio/http.html and we'll put something there to explain our messaging abstraction.

xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
```diff
@@ -1,14 +1,35 @@
...
+  String version();
+
+  Headers headers();
+
+  /**
+   * See: <a href=https://tools.ietf.org/html/rfc7540#section-5>rfc</a>
+   *
+   * @return the stream id of the http2 connection stream or {@link #H1_STREAM_ID_NONE} if http1
+   */
+  int streamId();
```
pdex: You default this in `Request` and `Response` we can just default it here.

xio-core/src/main/java/com/xjeffrose/xio/http/StreamingResponseData.java
```diff
@@ -5,6 +5,7 @@
 import io.netty.handler.codec.http.HttpResponseStatus;
 import lombok.ToString;
 
+/** Interface representing a partial HTTP1/2 Request */
```
pdex: Let's say segmented instead of partial

copy paste error: Request should be Response

xio-core/src/main/java/com/xjeffrose/xio/http/Message.java
manimaul: `streamId` is only defaulted in http.Response (not http.Request)

PR #209, opened 2018-04-13
wk/taco 131 client h2h1 streamid

manimaul:
OK - so we need to preserve the "stream id" of proxied h2 requests when the upstream server is h1 so that the proxy can manage the message session per stream id when transitioning between h2 and h1.


manimaul:
closing after conversation with @pdex : 
* we may not want to hang the streamid from the downstream request on top of the upstream request
* we should explore a more defensive option for streamid by handing off the streamid in the proxy handler
* or alternatively introduce the idea of "proxied request"

PR #211, opened 2018-04-16
Wk/taco 131 proxy handle h2h1 streamid

manimaul:
* tracking request/response per streamId (rather than per channel)
* adding steamId property to `SegmentedData`
* testing interleaved requests on the same channel

please look at the following pieces with a hairy eyeball:
Http2MessageSession, 
Http2ServerCodecUnitTest.testInterleavedStreamingMessages(),
Http2ServerCodec


xio-core/src/main/java/com/xjeffrose/xio/http/Http2MessageSession.java
```diff
@@ -0,0 +1,113 @@
...
+  private Http2MessageSession() {}
+
+  public void onRequest(Request request) {
+    MessageMetaState initialRequest = streamIdRequests.get(request.streamId());
+    if (initialRequest == null) {
+      if (request.startOfMessage()) {
+        streamIdRequests.put(
+            request.streamId(), new MessageMetaState(request, request.endOfMessage()));
+      } else {
+        log.error("Received an h2 message segment without a startOfMessage - request: {}", request);
```
pdex: This wording is a little confusing, maybe something more like "Received h2 Request object without startOfMessage == true".

xio-core/src/main/java/com/xjeffrose/xio/http/Http2MessageSession.java
```diff
@@ -0,0 +1,113 @@
...
+    MessageMetaState initialRequest = streamIdRequests.get(streamId);
+    if (initialRequest != null) {
+      return initialRequest.request;
+    }
+    return null;
+  }
+
+  /**
+   * Check if the message session has completed, if so remove state and prepare for the next
+   * session.
```
pdex: copy/paste error: "prepare for the next session" is confusing since we can have multiple sessions at the same time.

khappucino:
ok

PR #216, opened 2018-04-19
Wk/taco 131 proxied h2 to h1 transitions

manimaul:
* these changes address proxy requests where the front is h2 and back is h1
* the first change is tracking the h2 request stream id and propagating it to the ProxyResponse
* the second change is queuing subsequent h2 stream ids until the first stream id completes
(this is not ideal - ideally we would make a new h1 connection)

### DIFF
--- a/xio-core/src/test/java/com/xjeffrose/xio/http/Http2To1ProxyRequestQueueTest.java
+++ b/xio-core/src/test/java/com/xjeffrose/xio/http/Http2To1ProxyRequestQueueTest.java
@@ -1,0 +1,178 @@
+package com.xjeffrose.xio.http;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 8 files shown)


xio-core/src/test/java/com/xjeffrose/xio/http/Http2To1ProxyRequestQueueTest.java
```diff
@@ -0,0 +1,178 @@
...
+    // and the stream id should be correct
+    assertEquals(5, subject.currentProxiedH2StreamId().orElse(0), 0);
+
+    // and the the queue should be empty
+    assertTrue(subject.isEmpty());
+  }
+
+  @Test
+  public void testInterleavedH2FrontRequestsSegmentedResponse_stateTransitions() {
+    // should start empty
```
manimaul: yeah this test is too large - i'm not sure how else to cover all the state transitions

PR #218, opened 2018-04-20
TACO-2 added persistent proxy handler

jguamie:


xio-core/src/main/java/com/xjeffrose/xio/http/PersistentProxyHandler.java
```diff
@@ -0,0 +1,55 @@
...
+
+  @Override
+  public ClientConfig getClientConfig(Request request) {
+    String rawXFF = request.headers().get(X_FORWARDED_FOR).toString();
+    String remoteAddress = rawXFF;
+    if (rawXFF.contains(",")) {
+      // extract originating address
+      remoteAddress = rawXFF.substring(0, rawXFF.indexOf(","));
+    } else if (rawXFF == null) {
+      remoteAddress = request.host();
```
pdex: This isn't what you want. `Request.host` returns the host that the client is trying to connect to. HTTP/1.1 calls this a host header, h2 tried to clarify by referring to it as an authority. We should probably come up with a less confusing name as well.

The base class has some code for updating X_FORWARDED_FOR, I'm wondering if we should put similar logic in there to get the client address back out:
https://github.com/jguamie/xio/blob/384a49d334beeb5125f1186c8826946f860ab141/xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java#L93-L104

jguamie: Ok, will update.

xio-core/src/main/java/com/xjeffrose/xio/http/PersistentProxyHandler.java
```diff
@@ -0,0 +1,55 @@
...
+import io.netty.util.internal.PlatformDependent;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import lombok.val;
+
+public class PersistentProxyHandler extends ProxyHandler {
+  private final RendezvousHash<CharSequence> persistentProxyHasher;
+  private final Map<String, ClientConfig> clientConfigMap = PlatformDependent.newConcurrentHashMap();
```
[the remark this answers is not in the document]
jguamie: 👍

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
```diff
@@ -10,7 +10,7 @@
 
 @Slf4j
 public class ProxyHandler implements PipelineRequestHandler {
-  private static final AsciiString X_FORWARDED_FOR = AsciiString.cached("x-forwarded-for");
+  protected static final AsciiString X_FORWARDED_FOR = AsciiString.cached("x-forwarded-for");
```
pdex: this can go back to private.

jguamie: Done.

PR #220, opened 2018-04-24
TACO-35 adding support for multiple h2 client connection messages

manimaul:


xio-core/src/main/java/com/xjeffrose/xio/http/internal/SegmentedHttp2Response.java
```diff
@@ -40,7 +40,7 @@ public int streamId() {
 
   @Override
   public boolean startOfMessage() {
-    return false;
+    return true;
```
pdex: Whoops! Good catch.

PR #226, opened 2018-05-01
Setup gradle-release-plugin to play nice with bintray

pdex:


build.gradle
```diff
@@ -1,16 +1,39 @@
 plugins {
   id 'java'
+  id 'net.researchgate.release' version '2.7.0'
```
manimaul: these changes might be better in the xio build.gradle - `xio-core/build.gradle`

build.gradle
```diff
@@ -1,16 +1,39 @@
...
+    println "bintray_apikey: $bintray_apikey"
+    println "bintray_dryrun: ${bintray_dryrun}"
+  }
+}
+
+task imWithStupid(type:Exec) {
+  commandLine './gradlew', '-info', 'bintrayUpload'
+}
+
+afterReleaseBuild.dependsOn imWithStupid
```
pdex: There's something stupid going on with the `gradle-bintray-plugin` if I depend on `bintrayUpload` gradle says it can't find the tasks dependencies and freaks out. So what's happening here is that I'm shelling out after the tag is made and the binaries are built to upload them to bintray. If we wait until after the release process is done we'll be on the snapshot, which is not what we want.

xio-core/build.gradle
```diff
@@ -158,3 +140,35 @@ sourceSets {
...
+}
+
+bintray {
+  user = bintray_user //this usually comes from gradle.properties file in ~/.gradle
+  key = bintray_apikey //this usually comes from gradle.properties file in ~/.gradle
+  publications = ['maven'] //this refers to publishing.publications.maven up above ^^^
+  dryRun = bintray_dryrun.toBoolean() //Whether to run this as dry-run, without deploying
+  publish = true //If version should be auto published after an upload
+  pkg {
+    userOrg = bintray_subject
```
manimaul: should we use environment variables instead ? - `System.getenv('XXX')`

xio-core/build.gradle
```diff
@@ -158,3 +140,35 @@ sourceSets {
...
+    name = bintray_package
+    version {
+      name = project.version //Bintray logical version name
+    }
+    licenses = ['Apache-2.0']
+    websiteUrl = 'https://github.com/xjdr/xio'
+    issueTrackerUrl = 'https://github.com/xjdr/xio/issues'
+    vcsUrl = 'https://github.com/xjdr/xio.git'
+  }
+}
```
pdex: I tried to get rid of all of this boilerplate by using `maven-publish` like so: https://github.com/ajoberstar/gradle-bintray#archived

But the ;publish=1 bintray hack isn't supported by gradle right now.

build.gradle
manimaul: @pdex - that's probably because the tasks don't exist a the root level - see my comment above about moving these changes into the xio build.gradle

build.gradle
pdex: don't we want to run this plugin for both `xio-core` and `xio-test`?

xio-core/build.gradle
pdex: I set this up as a project variable so if you do `./gradlew -Pbintray_dryrun=true bintrayUpload` it works as expected. My understanding with environment variables and gradle is that you need to explicitly pass them to tasks.

RELEASING.md
```diff
@@ -1,36 +1,237 @@
 
-# prepare the release
+# did you remember to setup your bintray credentials?
```
jguamie: I do find Brian's instructions on "Create Bintray.com Account and Get Access to nordstromoss" and "Setup Bintray Credentials" to be quite valuable. It was very helpful for me to follow. Can we please replace your "did you remember to setup your bintray credentials" with Brian's first two sections? https://confluence.nordstrom.net/display/ENGP/XIO%3A+How+to+Publish+New+Version

jguamie: Nvm, I'm deleting the above remark cuz it is too company-specific.

PR #229, opened 2018-05-05
Wk/taco 20 - client (connection) pool

manimaul:
adding client pool with configurable per host max size
fixing proxy request stream mapping


xio-core/src/main/java/com/xjeffrose/xio/http/ClientPool.java
```diff
@@ -0,0 +1,84 @@
...
+        .reduce(0, (i, n) -> i + n);
+  }
+
+  public Client acquire(ClientConfig config, Supplier<Client> clientSupplier) {
+    return getPool(config.remote())
+        .values()
+        .stream()
+        .filter(
+            meta -> {
+              boolean available = meta.available.getAndSet(false);
```
manimaul: `getAndSet()` uses a spinlock

manimaul: > ConcurrentHashMaps support a set of sequential and parallel bulk operations that, unlike most > Stream methods, are designed to be safely, and often sensibly, applied even with maps that are > being concurrently updated by other threads; for example, when computing a snapshot > summary of the values in a shared registry.

https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/ConcurrentHashMap.html

PR #230, opened 2018-05-06
Dc/taco 151

khappucino:
I Modified ConfigReloader in the following ways:

ConfigReloader is now initialized with a filePath to the top level application.conf. However the ConfigReloader only updates if specific watched files are modified on disk (and they have different content than before).  The interface to add new files to watch is the ConfigReloader.addWatchFile(File file) method. This method will yell at you if the file you specifically tried to watch doesn't exist.  

Every time you call addWatchFile we compute the lastModifiedDate and the file contents digest. This is added to a HashMap keyed off of the file path.  Periodically when the update method is called we iterate through that hashmap of paths and we compare the current version of the watched file metadata (the lastModifiedDate and the computed digest of the contents), if there is a change we will reload the entire config tree and we will also update the metadata of the watch file at that path.

The way that you add an included file in the application config is like this:
application.conf =   trivial {  include "includedFile.conf" }
includedFile.conf = {  limit = "10" }

In the future we can have other included files and the addWatch method can be used to watch changes in those files as well. Any time there is a change in the dependency tree for the top level conf we recompute the entire configuration. 

Interesting findings***    If you provide an absolute path when you call ConfigFactory.parseFile(File file) and your top level file has a reference to another conf file in the same directory, the ConfigFactory.load(File file) method is smart enough to check the other files at the relative path in order to resolve the includedFile.conf file.

However, if you use ConfigFactory.parseReader(Reader reader), the ConfigFactory annoyingly only searches for the includedFile.conf in the existing classpath vs the same relative path as the top level conf.  This is why you see me adding the temp directories to the class path at run time.  This isn't going to be a problem in production because we specify the classpath when we run the nlp/nfe/xio deriatives, however during tests this was not the case.

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
@@ -13,6 +13,7 @@ import java.io.InputStreamReader;
 import java.io.Reader;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.*;
 import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -37,11 +38,23 @@ public class ConfigReloader<T> {
     }
   }
 
+  private static class ConfigFileMetadata {
+    final String path;
+    final long lastModified;
+    final byte[] digest;
+
+    ConfigFileMetadata(String path, long lastModified, byte[] digest) {
+      this.path = path;
+      this.lastModified = lastModified;
+      this.digest = digest;
+    }
+  }
+
   private final ScheduledExecutorService executor;
   private final Function<Config, T> factory;
   private BiConsumer<T, T> updater;
   private Meta<T> metadata;
-  private Map<String, Long> watchFiles = new HashMap<String, Long>();
+  private Map<String, ConfigFileMetadata> watchFiles = new HashMap<String, ConfigFileMetadata>();
 
   public ConfigReloader(ScheduledExecutorService executor, Function<Config, T> factory) {
     this.executor = executor;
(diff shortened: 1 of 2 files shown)


xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
```diff
@@ -72,24 +87,60 @@ private Config parse(Reader reader) {
     return load(new File(file));
   }
 
-  private boolean performUpdate(Meta<T> oldValue, Meta<T> newValue) {
-    return !MessageDigest.isEqual(oldValue.digest, newValue.digest);
+  private boolean performUpdate(ConfigFileMetadata current, ConfigFileMetadata previous) {
```
manimaul: these new parameter names are not as clear (and swapped with previous impl) - `oldValue` / `newValue`

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
```diff
@@ -72,24 +87,60 @@ private Config parse(Reader reader) {
...
+    return (current.lastModified > previous.lastModified
+        && !MessageDigest.isEqual(current.digest, previous.digest));
+  }
+
+  private ConfigFileMetadata loadMetaData(File configFile) throws IllegalStateException {
+    try {
+      MessageDigest digest = MessageDigest.getInstance("SHA-256");
+      DigestInputStream digester = new DigestInputStream(new FileInputStream(configFile), digest);
```
manimaul: i see a lot of `Closable` resources that are not being closed - these InputStreams and Readers should be closed ... consider using `try with resources`

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
```diff
@@ -72,24 +87,60 @@ private Config parse(Reader reader) {
...
+    } catch (NoSuchAlgorithmException e) {
+      throw new IllegalStateException(
+          "Illegal digest algorithm used on file '" + configFile.getAbsolutePath() + "'", e);
+    }
+  }
+
+  private boolean haveWatchFilesChanged() {
+    Boolean filesHaveChanged = false;
+    try {
+      Set<String> keys = new HashSet<String>(watchFiles.keySet());
```
manimaul: "nice to have": java streams api would be "nicer" here

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
```diff
@@ -106,6 +157,15 @@ void setUpdater(BiConsumer<T, T> updater) {
     this.updater = updater;
   }
 
+  public void addWatchFile(File file) throws IllegalStateException {
+    if (!watchFiles.containsKey(file.getAbsolutePath())) {
```
manimaul: this if check (whole function rather) isn't really necessary ... this isn't a MultiMap so only one value per key

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
manimaul: eg
```java
List<ConfigFileMetadata> toUpdateList = watchFiles.keySet().stream()
      .map(filePath -> loadMetaData(new File(filePath)))
      .filter(md -> {
        ConfigFileMetadata previousConfigMetadata = watchFiles.get(md.path);
        return performUpdate(md, previousConfigMetadata);
      })
      .collect(Collectors.toList());

    toUpdateList.forEach(update -> watchFiles.put(update.path, update));

    return !toUpdateList.isEmpty();
```

xio-core/src/test/java/com/xjeffrose/xio/config/ConfigReloaderUnitTest.java
```diff
@@ -48,122 +65,198 @@ String createConfig(String value) throws IOException, FileNotFoundException {
...
-    assertEquals(10, config.limit);
-    reloader.start(state::update);
-
-    Thread.sleep(2000);
+  private static void addPath(String s) throws Exception {
+    File f = new File(s);
+    URI u = f.toURI();
+    URLClassLoader urlClassLoader = (URLClassLoader) ClassLoader.getSystemClassLoader();
+    Class<URLClassLoader> urlClass = URLClassLoader.class;
+    Method method = urlClass.getDeclaredMethod("addURL", new Class[] {URL.class});
```
manimaul: line 79 and 81 - array is unnecessary for varargs

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
khappucino: ok i can change the name

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
khappucino: ok will do

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
khappucino: k

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
khappucino: name changed.

xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
khappucino: swapped to try with resources =p

xio-core/src/test/java/com/xjeffrose/xio/config/ConfigReloaderUnitTest.java
khappucino: ok changed

PR #237, opened 2018-05-14
Dc/taco 151 int test backend server

khappucino:
Initial work to support integration tests that require autoscaling groups and identifying the origin hosts that we hit.  ami-c00072b8 is an AMI that has a mockwebserver that the integration tests can hit


int-test-backend-server/scripts/spinup.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
```
nordjc: Maybe a comment here with an example of values to pass for things like `security-groups`?

int-test-backend-server/src/main/java/com/xjeffrose/xio/backend/server/Main.java
```diff
@@ -22,13 +24,13 @@
 public class Main {
 
   public static void main(String args[]) throws Exception {
+    if (args.length < 3) {
+      throw new RuntimeException("please specify server 'host' and 'port' and 'header-tag' arguments");
```
nordjc: Can we get an example of proper values to pass for `header-tag`?

PR #240, opened 2018-05-16
Wk/taco 231 proxy int test

manimaul:
testing the reverse proxy using python running the proxy and server(s) in separate processes

* int-test-proxy-server loads parameterized config
* added first pyunit reverse proxy test


int-tests/server_controller.py
```diff
@@ -1,21 +1,49 @@
...
+from queue import Queue, Empty
 
-_module_dir = os.path.abspath(os.path.join(os.path.dirname(__file__)))
-_root_dir = os.path.abspath(os.path.join(_module_dir, '..'))
+module_dir = path.abspath(path.join(path.dirname(__file__)))
+_root_dir = path.abspath(path.join(module_dir, '..'))
 
 
+class StdOutReader:
+  def __init__(self, stream, verbose=False):
```
manimaul: we needed a nonblocking reader to know when the server is ready to accept connections

int-tests/server_controller.py
```diff
@@ -1,21 +1,49 @@
...
+
+  def readline(self):
+    try:
+      line = str(self._queue.get(block=False, timeout=0.1))
+      if line and self._verbose:
+        print(line)
+      return line
+    except Empty:
+      return ''
+
```
khappucino: maybe we think of using healthchecks to see if the test servers are up in the futgure?

PR #242, opened 2018-05-17
Wk/taco 231 h2 int tests n stuff

manimaul:
* adding h2 integration tests with pycurl
* test speedup (building gradle artifacts once)
* configurable, full server process log output


int-tests/test_server.py
```diff
@@ -1,10 +1,19 @@
...
 
-class TestReverseProxyServer(unittest.TestCase):
+class TestReverseProxyServer(TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    global back_init
+    back_init = Initializer('int-test-backend-server')
+    global front_init
+    front_init = Initializer('int-test-proxy-server')
```
manimaul: even though Gradle has incremental builds - there's no need to invoke the artifact build process more than once

PR #245, opened 2018-05-17
dc/makeClientPoolMoreBetter added change so that we update the channe…

khappucino:
dc/makeClientPoolMoreBetter added change so that we update the channlAttr when we pull a client out of the clientpool as well as when we create a new client

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/ProxyClientFactory.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/ProxyClientFactory.java
@@ -29,7 +29,10 @@ public class ProxyClientFactory extends ClientFactory {
 
   @Override
   public Client getClient(ChannelHandlerContext ctx, ClientConfig config) {
-    return getHandlerClient(ctx)
-        .orElse(clientPool.acquire(config, () -> super.getClient(ctx, config)));
+    Client client =
+        getHandlerClient(ctx)
+            .orElse(clientPool.acquire(config, () -> super.getClient(ctx, config)));
+    updateChannelAttr(ctx, client);
+    return client;
   }
 }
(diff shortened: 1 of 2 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/ProxyClientFactory.java
```diff
@@ -29,7 +29,10 @@ public Client createClient(ChannelHandlerContext ctx, ClientConfig config) {
 
   @Override
   public Client getClient(ChannelHandlerContext ctx, ClientConfig config) {
-    return getHandlerClient(ctx)
-        .orElse(clientPool.acquire(config, () -> super.getClient(ctx, config)));
+    Client client =
+        getHandlerClient(ctx)
+            .orElse(clientPool.acquire(config, () -> super.getClient(ctx, config)));
+    updateChannelAttr(ctx, client);
```
manimaul: nice catch

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyClientFactory.java
```diff
@@ -29,7 +29,10 @@ public Client createClient(ChannelHandlerContext ctx, ClientConfig config) {
 
   @Override
   public Client getClient(ChannelHandlerContext ctx, ClientConfig config) {
-    return getHandlerClient(ctx)
-        .orElse(clientPool.acquire(config, () -> super.getClient(ctx, config)));
+    Client client =
+        getHandlerClient(ctx)
+            .orElse(clientPool.acquire(config, () -> super.getClient(ctx, config)));
```
manimaul: you could call `createClient` here rather than `super.getClient`  since we're always updating the channel attribute now

PR #247, opened 2018-05-19
Wk/taco 231 fix client pool

manimaul:
* fixed the client pool by swapping out the channel handler associated with the front-end when reusing a recycled client

There were 2 problems:
1 (found by @khappucino ) clients acquired by the pool were not being attached to the channel handler ctx
2 the client "app handler" (ProxyBackendHandler) was associated with the original front-end channel on pooled clients

### DIFF
--- a/int-test-proxy-server/src/main/resources/logback.groovy
+++ b/int-test-proxy-server/src/main/resources/logback.groovy
@@ -19,8 +19,10 @@ logger("io.netty.util.internal.PlatformDependent0", OFF)
 logger("io.netty.handler.ssl.CipherSuiteConverter", OFF)
 
 
-if (System.getProperty("COVERAGE") != null) {
+if (System.getProperty("DEBUG") != null) {
+  root(DEBUG, ["CONSOLE"])
+} else if (System.getProperty("COVERAGE") != null) {
   root(DEBUG, ["DEVNULL"])
 } else {
-  root(DEBUG, ["CONSOLE"])
+  root(WARN, ["CONSOLE"])
 }
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
@@ -41,7 +41,11 @@ public class Client {
             }
           }
         };
-    releaseListener = f -> channel = null;
+    releaseListener =
+        f -> {
+          log.debug("Channel closed");
+          channel = null;
+        };
   }
 
   public InetSocketAddress remoteAddress() {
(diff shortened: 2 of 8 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/ClientPool.java
```diff
@@ -27,20 +28,19 @@ public ClientPool(int maxSizePerAddress) {
...
-    //      log.debug("releasing client to pool {}", client);
-    //      pool.put(client, new Meta(client));
-    //    } else {
-    //      Meta meta = pool.get(client);
-    //      if (meta != null) {
-    //        log.debug("setting client available in pool {}", client);
-    //        meta.available.set(true);
-    //      }
-    //    }
+    log.debug("recycling client {}", client);
```
manimaul: there is no real change here - other than uncommenting

int-test-proxy-server/src/main/resources/logback.groovy
```diff
@@ -19,8 +19,10 @@ logger("io.netty.util.internal.PlatformDependent0", OFF)
...
 
 
-if (System.getProperty("COVERAGE") != null) {
+if (System.getProperty("DEBUG") != null) {
+  root(DEBUG, ["CONSOLE"])
+} else if (System.getProperty("COVERAGE") != null) {
   root(DEBUG, ["DEVNULL"])
 } else {
-  root(DEBUG, ["CONSOLE"])
+  root(WARN, ["CONSOLE"])
```
manimaul: reducing verbosity of the ReverseProxyFunctionalTest

xio-core/src/main/java/com/xjeffrose/xio/http/ClientPool.java
```diff
@@ -65,6 +66,7 @@ public Client acquire(ClientConfig config, Supplier<Client> clientSupplier) {
               if (available) {
                 int count = meta.usageCount.incrementAndGet();
                 log.debug("reusing client in pool with usage count {}", count);
+                meta.client.prepareForReuse(() -> new ProxyBackendHandler(ctx));
```
manimaul: here is the fix to the second issue - replacing the "app handler" associated with the appropriate "front end"

xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -78,11 +82,26 @@ public ChannelFuture write(Request request) {
...
+      if (request.endOfMessage()) {
+        return channel.writeAndFlush(request).addListener(writeListener);
+      } else {
+        return channel.write(request).addListener(writeListener);
+      }
+    }
+  }
+
+  public void prepareForReuse(Supplier<ChannelHandler> handlerSupplier) {
+    clientChannelInitializer.setAppHandler(handlerSupplier);
```
pdex: FYI, If we're reusing an open connection the channel initializer isn't going to be invoked.

PR #252, opened 2018-05-30
Make the acquiring of a client from the pool only happen when getHand…

codeman9:
…lerClient is null


xio-core/src/main/java/com/xjeffrose/xio/http/ProxyClientFactory.java
```diff
@@ -31,7 +31,7 @@ public Client createClient(ChannelHandlerContext ctx, ClientConfig config) {
   public Client getClient(ChannelHandlerContext ctx, ClientConfig config) {
     Client client =
         getHandlerClient(ctx)
-            .orElse(clientPool.acquire(ctx, config, () -> createClient(ctx, config)));
+            .orElseGet(() -> clientPool.acquire(ctx, config, () -> createClient(ctx, config)));
```
manimaul: nice catch

PR #253, opened 2018-05-31
TACO-309 containerize an h2load test

manimaul:
* pyunit h2load test
* adding `Dockerfile`s for the int-test services and pyunit test client
* adding docker-compose for running  load test

### DIFF
--- a/int-tests/Dockerfile
+++ b/int-tests/Dockerfile
@@ -1,0 +1,7 @@
+FROM debian:sid
+ADD ./*.py /tests/
+ADD ./requirements.txt /tests/requirements.txt
+RUN apt-get update
+RUN apt-get -y upgrade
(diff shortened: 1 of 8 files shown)


int-tests/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM debian:sid
+ADD ./*.py /tests/
+ADD ./requirements.txt /tests/requirements.txt
+RUN apt-get update
+RUN apt-get -y upgrade
+RUN apt-get install -y python3.7 python3-pip python3-pycurl nghttp2-client wrk
```
codeman9: python 3.7?

manimaul: is suppose this s/could be python 3.6 ¯\_(ツ)_/¯

PR #259, opened 2018-06-06
Dc/taco 326

khappucino:
We stop closing the frontEnd just because the client channel is idle

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/core/XioIdleDisconnectHandler.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/core/XioIdleDisconnectHandler.java
@@ -1,6 +1,5 @@
 package com.xjeffrose.xio.core;
 
-import com.xjeffrose.xio.http.internal.ProxyClientIdle;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.handler.timeout.IdleStateEvent;
 import io.netty.handler.timeout.IdleStateHandler;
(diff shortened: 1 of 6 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/internal/ProxyClientIdle.java
```diff
@@ -0,0 +1,7 @@
+package com.xjeffrose.xio.http.internal;
+
+public final class ProxyClientIdle {
```
manimaul: an enum might be more appropriate

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyBackendHandler.java
```diff
@@ -54,7 +55,7 @@ public void channelInactive(ChannelHandlerContext ctx) throws Exception {
     // TODO(CK): this should really be some sort of notification to the frontend
     // that the backend closed. Keepalive/h2 will require the connection to stay open, we
     // shouldn't be closing it.
-    frontend.close();
+    frontend.fireUserEventTriggered(ProxyClientIdle.INSTANCE);
```
pdex: This isn’t idle so much as closed. 

https://netty.io/4.1/api/io/netty/channel/ChannelInboundHandler.html#channelInactive-io.netty.channel.ChannelHandlerContext-

xio-core/src/main/java/com/xjeffrose/xio/core/XioIdleDisconnectHandler.java
```diff
@@ -26,6 +27,14 @@ public XioIdleDisconnectHandler(
         TimeUnit.MILLISECONDS);
   }
 
+  @Override
+  public void userEventTriggered(ChannelHandlerContext ctx, Object evt) throws Exception {
+    super.userEventTriggered(ctx, evt);
+    if (evt instanceof ProxyClientIdle) {
+      ctx.channel().close();
```
pdex: This isn’t really in the spirit of what this handler is supposed to be doing. What we probably want is to notify the business logic around the proxy client that things went south.

khappucino:
screen shot 2018-06-07 at 11 52 56 am

khappucino:
Punting for now

PR #260, opened 2018-06-07
TACO-326-supreme removing the idle channel handler from the ClientCha…

khappucino:
…nnelInitializer so that we don't manually close the channel just because it is idling, we are trying to keep the client channels open so when we reuse the client we don't have to do the whole handshake over again to connect

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
@@ -14,12 +14,14 @@ import lombok.extern.slf4j.Slf4j;
 @Accessors(fluent = true)
 @Getter
 public class ClientConfig {
+
   private final Map<ChannelOption<Object>, Object> bootstrapOptions;
   private final String name;
   private final TlsConfig tls;
   private final boolean messageLoggerEnabled;
   private final InetSocketAddress local;
   private final InetSocketAddress remote;
+  private final IdleTimeoutConfig idleTimeoutConfig;
 
   public String getName() {
     return name;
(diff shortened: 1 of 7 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/ClientChannelInitializer.java
```diff
@@ -64,7 +63,6 @@ protected void initChannel(Channel channel) throws Exception {
     }
     channel
         .pipeline()
-        .addLast("idle handler", new XioIdleDisconnectHandler(60, 60, 60))
```
manimaul: i think client is meant to be used not just as a proxy client... i wonder if the disconnect handler should be removed by the proxy client factory instead

manimaul: ... or better yet a configurable timeout

khappucino: yeah making update, good idea

khappucino: I will add tests as well

xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
```diff
@@ -14,12 +14,14 @@
...
 @Getter
 public class ClientConfig {
+
   private final Map<ChannelOption<Object>, Object> bootstrapOptions;
   private final String name;
   private final TlsConfig tls;
   private final boolean messageLoggerEnabled;
   private final InetSocketAddress local;
   private final InetSocketAddress remote;
+  private final IdleTimeoutConfig idleTimeoutConfig;
```
pdex: Like!

xio-core/src/main/resources/reference.conf
```diff
@@ -163,6 +163,8 @@ xio {
     remotePort = 0
     localIp = ""
     localPort = 0
+    idleTimeoutEnabled = false
+    idleTimeoutDuration = 0
```
pdex: At some point we may want the type of incoming connection to determine if there’s no idle timeout. This is fine for now.

khappucino: k

PR #261, opened 2018-06-20
Br/taco 340

Rivukis:
This is the handler that apps (NFE, etc.) will use to make gRPC handlers. This does all the heavy so users only specify the Request Type, Response Type, and business logic

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/GrpcRequestHandler.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/GrpcRequestHandler.java
@@ -71,14 +71,16 @@ public class GrpcRequestHandler<GrpcRequest extends
         return;
       }
 
-      ByteBuf contentBuffer = UnpooledByteBufAllocator.DEFAULT.buffer(size, size);
       int packetContentSize = actualBuffer.readableBytes() - 5;
+      ByteBuf contentBuffer = UnpooledByteBufAllocator.DEFAULT.buffer(packetContentSize, size);
       contentBuffer.writeBytes(actualBuffer.slice(5, packetContentSize));
 
       state = new GrpcState(size, contentBuffer);
     } else {
       state.buffer.writeBytes(actualBuffer);
     }
+
+    session.put(request.streamId(), state);
 
     if (request.endOfMessage()) {
       if (state.size != state.buffer.readableBytes()) {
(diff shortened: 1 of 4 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/GrpcRequestHandler.java
```diff
@@ -0,0 +1,283 @@
...
+  public void handle(ChannelHandlerContext ctx, Request request, RouteState route) {
+    ByteBuf actualBuffer;
+    if (request instanceof SegmentedRequestData) {
+      SegmentedRequestData segmentedRequest = (SegmentedRequestData) request;
+      actualBuffer = segmentedRequest.content();
+    } else {
+      actualBuffer = request.body();
+    }
+
+    HashMap<Integer, GrpcState> session = lazyCreateSession(ctx);
```
codeman9: ⚪️ If actualBuffer is null or not readable we will still execute these 2 lines...Does that matter?

xio-core/src/main/java/com/xjeffrose/xio/http/GrpcRequestHandler.java
```diff
@@ -0,0 +1,283 @@
...
+    GrpcState state = session.get(request.streamId());
+
+    if (actualBuffer == null || !actualBuffer.isReadable()) {
+      return;
+    }
+
+    if (state == null) {
+      boolean haveMetaData = actualBuffer.isReadable(5);
```
codeman9: ⚪️ Would it be beneficial to have the '5' as a constant somewhere? I see it in a few places.

xio-core/src/main/java/com/xjeffrose/xio/http/GrpcRequestHandler.java
Rivukis: it won't change the logic, but I'll move them down.

xio-core/src/main/java/com/xjeffrose/xio/http/GrpcRequestHandler.java
Rivukis: fixed along with a few other clean ups

xio-core/src/test/java/com/xjeffrose/xio/http/GrpcRequestHandlerTest.java
```diff
@@ -0,0 +1,345 @@
...
+
+    channel =
+        new EmbeddedChannel(
+            new SimpleChannelInboundHandler<Request>() {
+              @Override
+              protected void channelRead0(ChannelHandlerContext ctx, Request request) {
+                subject.handle(ctx, request, null);
+              }
+            });
+  }
```
pdex: I like this setup and how the tests are looking for the responses to come out of the handler. excellent!

PR #266, opened 2018-06-27
Br/taco 236

Rivukis:


Rivukis:
This undos the changes to RouteState to make it non-file friendly and moves that friendliness to RouteConfig to keep the pattern of mutable changes on top of immutable initial values (i.e. RouteState on top of RouteConfig)

PR #267, opened 2018-06-28
TACO-372 - remove content-length header when transfer encoding chunke…

codeman9:
…d header is set.


xio-core/src/test/java/com/xjeffrose/xio/http/EdgeProxyFunctionalTest.java
```diff
@@ -306,6 +308,9 @@ void post(String prefix, int port) throws Exception {
     server.enqueue(buildResponse());
     Response response = client.newCall(request).execute();
     assertEquals(200, response.code());
+    if (response.headers().names().contains(HttpHeaderNames.TRANSFER_ENCODING.toString())) {
```
Rivukis: if statement in tests?

codeman9: Removed if statement.

PR #270, opened 2018-06-30
TACO-346 updated the configreloader so that we compute the digest separately from reading the file

khappucino:
TACO-346 updated the configreloader so that we compute the digest separately from reading the file This in general would be good, however when we read the application.conf in this way, it cannot find the files included in the application.conf such as include 'route.conf'.  By switching to the Config.parseFile we maintain the relative path of the route.conf file and it doesn't need to be in the classpath, what I was seeing was that when we turn it into a stream and do a Config.parseReader, we lost context fo the original file location and Config would only look in the classpath to find any files included in the application.conf file

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
@@ -7,6 +7,8 @@ import com.google.common.annotations.VisibleForTesting;
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
 import java.io.*;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.security.DigestInputStream;
 import java.security.MessageDigest;
 import java.util.*;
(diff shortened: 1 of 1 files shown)


xio-core/src/main/java/com/xjeffrose/xio/config/ConfigReloader.java
```diff
@@ -58,35 +60,28 @@ public ConfigReloader(ScheduledExecutorService executor, Function<Config, T> fac
...
-      throw new IllegalStateException(
-          "Couldn't load config from file '" + file.getAbsolutePath() + "'", e);
+      throw new IllegalStateException("Couldn't load config from file '" + filePath + "'", e);
     }
-
-    try (FileInputStream fileInputStream = new FileInputStream(file);
-        DigestInputStream digestInputStream = new DigestInputStream(fileInputStream, digest);
-        InputStreamReader reader = new InputStreamReader(digestInputStream)) {
-      T value = factory.apply(parse(reader));
+    try {
```
Rivukis: try with resource will auto close the resources. regular try doesn't. Is that an issue here since you are changing it from try with resources to just a regular try?

khappucino: will double check if needed

khappucino: file does not implement closable so =p will have to close manually

PR #271, opened 2018-07-03
Wk/taco 378 preserve explicitly set transfer-encoding header

manimaul:
preserve explicitly set transfer-encoding header for proxied h2 to h1 requests

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/Http1ServerCodec.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/Http1ServerCodec.java
@@ -11,19 +11,7 @@ import io.netty.channel.ChannelFuture;
 import io.netty.channel.ChannelFutureListener;
 import io.netty.channel.ChannelHandlerContext;
 import io.netty.channel.ChannelPromise;
-import io.netty.handler.codec.http.DefaultFullHttpResponse;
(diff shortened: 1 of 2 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/Http1ServerCodec.java
```diff
@@ -134,18 +122,14 @@ private void buildResponse(ChannelHandlerContext ctx, Response response, Channel
                 full.headers().http1Headers(false, false),
                 EmptyHttpHeaders.INSTANCE);
       } else {
-        // TODO(CK): TransferEncoding
         // We don't know the size of the message payload so set TransferEncoding to chunked
+        HttpHeaders headers = response.headers().http1Headers(false, false);
```
manimaul: when `Headers` is an instance of `Http2HeadersWrapper` because of an h1 to h2 transition - transfer-encoding is stripped out by `http1Headers()` call ... so we make the transition up front

pdex: ahh nice and clean and my lazy hack gets to stay!

manimaul: laziness for the win

PR #273, opened 2018-07-05
TACO-379 - address intermittent concurrent NPE in RendezvousHash

manimaul:


xio-core/src/main/java/com/xjeffrose/xio/server/RendezvousHash.java
```diff
@@ -50,11 +52,11 @@ public T getOne(byte[] key) {
               hasher.newHasher().putBytes(key).putObject(node, nodeFunnel).hash().asLong(), node);
         });
 
-    return hashMap.remove(hashMap.keySet().stream().max(Long::compare).orElse(null));
+    return hashMap.keySet().stream().max(Long::compare).map(hashMap::remove).orElse(null);
```
pdex: can we doc this real quick?
```java
hashMap.keySet().stream() // iterate over keyset
.max(Long::compare) // find the largest key
.map(hashMap::remove) // remove the node from the map, return the node
.orElse(null); // return null if the map is empty
```

manimaul: ok - i also changed the remove to a get as the map is thrown away anyhow

PR #277, opened 2018-07-11
Dc/taco 394 enqueue requests before connect completes

khappucino:
make sure we connect before trying to write to the channel.
If you try to connect and write back to back the AbstractChannelContextHandler does the findOutboundContext check before the connection is completed (and the ClientChannelInitializer actually registers) and the first part of the message can be sent directly to the ChannelHEAD without going through any of the channel handlers.  The eventLoop thread checking happens AFTER the nextOutboundContext is stored.  We check the connection state to determine what course of action to take. This works well since the Client Channel is using the same event loop as the Server channel so we don't need to use concurrent queue etc... The new MR throwing out client's who are no reusable lets us sidestep the issue where if you wanted to reconnect on an old client, it's bootstrap is still tied to the previous server channel's eventloop.


khappucino:
h2load -c 5000 -t 8 -D 60 https://172.28.75.250:8888/v1/
finished in 60.36s, 25964.38 req/s, 1.86MB/s
requests: 1557863 total, 1562863 started, 1557863 done, 1557863 succeeded, 0 failed, 0 errored, 0 timeout
status codes: 1557863 2xx, 0 3xx, 0 4xx, 0 5xx
traffic: 111.78MB (117204725) total, 9.15MB (9595172) headers (space savings 93.08%), 68.36MB (71684652) data
                     min         max         mean         sd        +/- sd
time for request:      611us       7.70s    177.62ms    437.39ms    98.49%
time for connect:   489.00ms       5.53s       2.69s       1.04s    67.72%
time to 1st byte:      3.20s       8.06s       6.15s       1.08s    59.76%
req/s           :       2.84        6.46        5.19        0.56    73.98%



wrk -c 5000 -t 8 -d 60 https://172.28.75.250:8888/v1/
Running 1m test @ https://172.28.75.250:8888/v1/
  8 threads and 5000 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   157.77ms   91.87ms   2.00s    84.65%
    Req/Sec     3.97k     1.15k    5.46k    82.14%
  1704139 requests in 1.00m, 334.79MB read
  Socket errors: connect 0, read 0, write 0, timeout 4788
Requests/sec:  28363.67
Transfer/sec:      5.57MB

gradle.properties
```diff
@@ -1,5 +1,5 @@
 # project_version=0.13.5-SNAPSHOT
-version=0.13.25-SNAPSHOT
+version=0.13.25-SNAPSHOT-TACO-394
```
manimaul: i don't think you want this

khappucino: yeah sorry left over from int testing will remove

xio-core/src/main/java/com/xjeffrose/xio/http/ClientConnectionManager.java
```diff
@@ -0,0 +1,82 @@
...
+      return currentChannelFuture;
+    }
+  }
+
+  public ClientConnectionState connectionState() {
+    return connectionState;
+  }
+
+  public boolean isReusable() {
+    return reusable;
```
manimaul: the reusable member variable and this method could go away... just check if connectionState != CLOSED...

xio-core/src/test/java/com/xjeffrose/xio/http/ClientConnectionManagerIntegrationTest.java
```diff
@@ -0,0 +1,95 @@
...
+    assertEquals(subject.connectionState(), ClientConnectionState.NOT_CONNECTED);
+  }
+
+  @Test
+  public void testConnectingFailingConnection() throws Exception {
+    subject = subjectFactory(false);
+    // don't set up fake origin backend server so we can connect to it
+    Future<Void> connectionResult = subject.connect();
+    assertEquals(ClientConnectionState.CONNECTING, subject.connectionState());
+    connectionResult.awaitUninterruptibly(60000);
```
manimaul: 60 seconds is in the client configuration for conn timeout ... maybe lower it in the test config

xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -78,39 +50,103 @@ public ChannelFuture connect() {
...
+  }
+
+  private void executeBufferedRequests(Future<? super Void> connectionResult) {
+    boolean connectionSuccess = connectionResult.isDone() && connectionResult.isSuccess();
+    log.debug("== Connection success was " + connectionSuccess);
+    // loop through the queue until it's empty and fire away
+    // this will happen on the same eventloop as the write so we don't need to worry about
+    // trying to write to this queue at the same time we are dequeing
+    while (!requestQueue.isEmpty()) {
+      Client.ClientPayload pair = (Client.ClientPayload) requestQueue.remove();
```
Rivukis: Is `pair` a normal word to describe a `ClientPayload`?

xio-core/src/main/java/com/xjeffrose/xio/http/ClientConnectionManager.java
```diff
@@ -0,0 +1,82 @@
...
+
+    releaseListener =
+        f -> {
+          log.debug("Channel closed");
+          connectionState = ClientConnectionState.CLOSED_CONNECTION;
+          this.currentChannelFuture = null;
+          this.reusable = false;
+        };
+    connectionListener =
+        f -> {
```
Rivukis: Do you mind changing the variable name to be more descriptive than a single letter?

Rivukis: Same question a few lines above as well

xio-core/src/test/java/com/xjeffrose/xio/http/ClientConnectionManagerIntegrationTest.java
khappucino: yeah so, it takes up to that long for the connection to time out, we could configure the client pipeline to have a configurable timeout, but that should be done in another ticket since that is non-zero amount of work.

khappucino: I will try an invalidly formed ip address to see if i can get the channel connect to fail faster

khappucino: ok using a bogus ip address lets us fail the connection faster, will update pr

khappucino: updated

xio-core/src/main/java/com/xjeffrose/xio/http/ClientConnectionManager.java
khappucino: ok i can do that

khappucino: ok so the isReusable is now return manager.connectionState == ClientConnectionState.CONNECTED, we shouldn't have NOT_CONNECTED, CONNECTING or CLOSED_CONNECTION clients in the pool, only CONNECTED =p

xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
khappucino: I can rename it, i really just wish we had tuple =p

khappucino: ok updated it to requestPayload instead of pair

PR #278, opened 2018-07-11
zk exhibitor integration

pdex:
This PR sets up the `Application` class's `ZkClient` to be used with an exhibitor ensemble.

  * If `settings.zookeeper.cluster` starts with "ensemble" the string will be parsed with this assumed format: "ensemble:<port>:<server1>,<server2>,<serverN>"
  * A simple app is included to test with a proper exhibitor ensemble
  * A flaky test was also fixed
  * verifyGoogleJavaFormat now ignores more generate sources

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/application/ApplicationConfig.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/application/ApplicationConfig.java
@@ -10,6 +10,8 @@ import com.xjeffrose.xio.core.NullZkClient;
 import com.xjeffrose.xio.core.ZkClient;
 import com.xjeffrose.xio.tracing.XioTracing;
 import io.netty.util.internal.PlatformDependent;
+import java.util.Arrays;
+import java.util.Collection;
 import java.util.List;
 import java.util.Map;
 import java.util.function.Function;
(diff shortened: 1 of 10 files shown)


xio-core/src/main/java/com/xjeffrose/xio/application/ApplicationConfig.java
```diff
@@ -104,11 +106,19 @@ public ServerChannelConfiguration serverChannelConfig() {
...
 
+  // TODO(CK): refactor this into a ZooKeeperConfig that the ZkClient constructor will accept
   public ZkClient zookeeperClient() {
     if (zookeeperCluster.isEmpty()) {
       return new NullZkClient();
     } else {
-      return new ZkClient(zookeeperCluster);
+      if (zookeeperCluster.startsWith("exhibitor:")) {
+        String[] values = zookeeperCluster.replace("exhibitor:", "").split(":");
+        int restPort = Integer.parseInt(values[0]);
```
manimaul: maybe bounds check array before indexing into it

pdex: I'm fine with this throwing an exception if you did something silly. We can clean this up in a ZooKeeperConfig class later.

PR #279, opened 2018-07-12
Br/taco 385 - adding builders for ProxyRouteConfig and ClientConfig to make these at runtime

Rivukis:


xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
```diff
@@ -74,4 +92,78 @@ public static ClientConfig fromConfig(String key, Config config) {
...
+    return new ClientConfig.Builder(fallbackObject);
+  }
+
+  /**
+   * Used to create a ClientConfig at runtime.
+   *
+   * <p>If a value is not set, it defaults to using the default object's value.
+   */
+  public static class Builder {
+    private ClientConfig fallbackObject;
```
pdex: The name is fallbackObject the javadoc says default object. I would pick one name and stick with it.

xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
```diff
@@ -74,4 +92,78 @@ public static ClientConfig fromConfig(String key, Config config) {
...
+    private Map<ChannelOption<Object>, Object> bootstrapOptions;
+    private String name;
+    private TlsConfig tls;
+    private boolean messageLoggerEnabled;
+    private InetSocketAddress local;
+    private InetSocketAddress remote;
+    private IdleTimeoutConfig idleTimeoutConfig;
+
+    private Builder(ClientConfig fallbackObject) {
```
Rivukis: This is a ClientConfig instead of a Config so that the user has the capabilities to create a default object once and use it over and over instead of newing up one everytime with the same Config

Rivukis: Same reasoning in ProxyRouteConfig as well

xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
```diff
@@ -74,4 +92,78 @@ public static ClientConfig fromConfig(String key, Config config) {
...
+      this.fallbackObject = fallbackObject;
+    }
+
+    public Builder setBootstrapOptions(Map<ChannelOption<Object>, Object> bootstrapOptions) {
```
pdex: If I was writing this from scratch i’d use a google auto builder. Don’t change anything here. Just look into it for next time so you don’t have to create a bunch of boilerplate. Then again. I’m hand crafting code with sed and you’re using intellij so i’m sure this was painless.

xio-core/src/test/java/com/xjeffrose/xio/client/ClientConfigTest.java
```diff
@@ -0,0 +1,54 @@
...
+import org.junit.Test;
+
+public class ClientConfigTest extends Assert {
+  @Test
+  public void testBuilderCustomValues() {
+    Config config = ConfigFactory.load();
+    ClientConfig fallbackObject = new ClientConfig(config.getConfig("xio.clientTemplate"));
+    Map expectedBootstrapOptions = Maps.newHashMap(ChannelOption.AUTO_READ, new Object());
+    String expectedName = "name";
+    TlsConfig expectedTls = new TlsConfig(ConfigFactory.load("tls-reference"));
```
Rivukis: I'm trying to avoid having to make a builder for TlsConfig for now in hopes that we won't need one. If we need to be able to create different ones at runtime, we can add it then

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyRouteConfig.java
```diff
@@ -64,4 +65,101 @@ public ProxyRouteConfig(Config config) {
...
+    return new Builder(fallbackObject);
+  }
+
+  /**
+   * Used to create a ProxyRouteConfig at runtime.
+   *
+   * <p>If a value is not set, it defaults to using the default object's value.
+   */
+  public static class Builder {
+    private ProxyRouteConfig fallbackObject;
```
pdex: Same note as before either default or fallback as the name.

xio-core/src/test/java/com/xjeffrose/xio/client/ClientConfigTest.java
pdex: You shouldn’t need a builder for TlsConfig. We’ll always want something typesafe config based so we can pull in defaults from reference.conf

xio-core/src/test/resources/application.conf
```diff
@@ -209,12 +209,6 @@ xio {
     name = "test client"
   }
 
-  proxyRouteTemplate = ${xio.routeTemplate} {
```
pdex: Nice. We missed this before.

xio-core/src/test/java/com/xjeffrose/xio/client/ClientConfigTest.java
Rivukis: That's what I was hoping for 👍

xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
Rivukis: fixed

xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
Rivukis: Oh no, it was painful. I just didn't know about auto builder. I'll look into it

xio-core/src/main/java/com/xjeffrose/xio/http/ProxyRouteConfig.java
Rivukis: fixed

xio-core/src/main/java/com/xjeffrose/xio/client/ClientConfig.java
Rivukis: looked into it. got close to making it work, but it's not compiling and the output is not showing me the error. I'm out of ideas so moving forward with my hand rolled version

PR #289, opened 2018-08-07
TACO-466: add XioServiceLocator and resolved circular dependency issu…

Rivukis:
…es during app initialization


xio-core/src/main/java/com/xjeffrose/xio/application/XioServiceLocator.java
```diff
@@ -0,0 +1,27 @@
+package com.xjeffrose.xio.application;
+
+import com.xjeffrose.xio.core.ZkClient;
+import lombok.Getter;
+
+public class XioServiceLocator {
+  public static XioServiceLocator INSTANCE = null;
```
manimaul: we should add some access control to the INSTANCE so that only the application root scope can initialize it
... maybe add a private, empty constructor and perhaps a  static factory constructor which takes in the application bootstrap

manimaul: or maybe move this over to the bootstrap package and make the initialization "package private" in order to clearly communicate that this is intended as an ioc container that is initialized only at the application root

Rivukis: fixed

xio-core/src/main/java/com/xjeffrose/xio/bootstrap/XioServiceLocator.java
```diff
@@ -0,0 +1,41 @@
...
+  private ApplicationConfig applicationConfig;
+  private ApplicationState applicationState;
+
+  private XioServiceLocator(ApplicationConfig applicationConfig, ApplicationState applicationState) {
+    this.applicationConfig = applicationConfig;
+    this.applicationState = applicationState;
+  }
+
+  public static XioServiceLocator getInstance() {
+    return instance;
```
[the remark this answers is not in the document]
Rivukis: added

xio-core/src/main/java/com/xjeffrose/xio/application/ApplicationConfig.java
```diff
@@ -36,6 +32,8 @@
   @Getter private final double hardReqPerSec;
   @Getter private final int rateLimiterPoolSize;
   @Getter private final int clientPoolSize;
+
+  // TODO(br): this should be moved out of here since ApplicationConfig should only hold onto immutable configuration info
```
pdex: agreed, this was a lazy move on my part :)

xio-core/src/main/java/com/xjeffrose/xio/bootstrap/ApplicationBootstrap.java
```diff
@@ -95,7 +98,14 @@ public ApplicationBootstrap addServer(
   public Application build() {
     Map<String, XioServer> servers = new HashMap<>();
     serverBootstraps.forEach((k, v) -> servers.put(k, v.build()));
+
+    ZkClient zkClient = state.getZkClient();
+    zkClient.registerUpdater(
+        new IpFilterConfig.Updater(config.getIpFilterPath(), state::setIpFilterConfig));
+    zkClient.registerUpdater(
+        new Http1FilterConfig.Updater(config.getHttp1FilterPath(), state::setHttp1FilterConfig));
```
pdex: nice, I like that this is out of the `ApplicationState`

PR #292, opened 2018-08-15
TACO-202 - Customize the OkHttpClient for Zipkin to not use compressi…

codeman9:
…on and to not perform ssl validation.

### DIFF
--- a/xio-core/src/main/java/zipkin2/reporter/okhttp3/OkHttpClientBuilderFactory.java
+++ b/xio-core/src/main/java/zipkin2/reporter/okhttp3/OkHttpClientBuilderFactory.java
@@ -1,0 +1,45 @@
+package zipkin2.reporter.okhttp3;
+
+import java.security.KeyManagementException;
+import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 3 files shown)


xio-core/src/main/java/zipkin2/reporter/okhttp3/OkHttpClientBuilderFactory.java
```diff
@@ -0,0 +1,45 @@
...
+    SSLContext sslCtx = null;
+    try {
+      sslCtx = SSLContext.getInstance("TLSv1.2");
+      sslCtx.init(null, trustManagers, null);
+    } catch (NoSuchAlgorithmException | KeyManagementException e) {
+      System.exit(-1);
+      throw new RuntimeException(e);
+    }
+
+    return new OkHttpClient.Builder()
```
khappucino: maybe later we can have this be a parameter from config file? to disable host checkin

codeman9: This sounds like a great idea. I will create a ticket.

PR #299, opened 2018-09-06
TACO-543 - add datadog support via an enum type in the configuration …

codeman9:
…file.

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/config/TracingConfig.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/config/TracingConfig.java
@@ -24,27 +24,23 @@ public class TracingConfig {
     }
     this.type = type;
 
-    switch (type) {
-      case ZIPKIN:
-        String zipkinUrl;
-        if (config.hasPath("zipkin.zipkinUrl")) {
-          zipkinUrl = config.getString("zipkin.zipkinUrl");
-        } else {
-          zipkinUrl = "";
-        }
+    if (type == TracingType.ZIPKIN) {
+      String zipkinUrl;
+      if (config.hasPath("zipkin.zipkinUrl")) {
+        zipkinUrl = config.getString("zipkin.zipkinUrl");
+      } else {
+        zipkinUrl = "";
+      }
 
-        float samplingRate;
-        if (config.hasPath("zipkin.samplingRate")) {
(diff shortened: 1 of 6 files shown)


xio-core/src/main/java/com/xjeffrose/xio/config/TracingConfig.java
```diff
@@ -3,20 +3,57 @@
...
+          samplingRate = ((Double) config.getDouble("zipkin.samplingRate")).floatValue();
+        } else {
+          samplingRate = 0.01f;
+        }
+
+        this.zipkinUrl = zipkinUrl;
+        this.zipkinSamplingRate = samplingRate;
+        break;
+      case DATADOG:
+        break;
```
manimaul: maybe log the configured tracer type

xio-core/src/main/java/com/xjeffrose/xio/config/TracingConfig.java
```diff
@@ -3,20 +3,57 @@
...
+    TracingType type;
+    if (config.hasPath("type")) {
+      type = config.getEnum(TracingType.class, "type");
+    } else {
+      type = TracingType.ZIPKIN;
+    }
+    this.type = type;
+
+    switch (type) {
+      case ZIPKIN:
```
manimaul: maybe `if (type == ZIPKIN)` rather than switch since we only care about 1 case

xio-core/src/main/java/com/xjeffrose/xio/tracing/XioTracing.java
```diff
@@ -23,11 +24,20 @@
...
     name = config.getApplicationName();
-    String zipkinUrl = config.getZipkinUrl();
-    float samplingRate = config.getZipkinSamplingRate();
-    Tracing tracing = buildZipkinTracing(this.name, zipkinUrl, samplingRate);
-    if (tracing != null) {
-      tracer = BraveTracer.create(tracing);
+    TracingConfig.TracingType type = config.getType();
+
+    switch (type) {
+      case ZIPKIN:
```
manimaul: maybe log the configured tracer type here

codeman9: now logging the tracer type as log.info

PR #303, opened 2018-09-14
AWS Zookeeper registration

pdex:


settings.gradle
```diff
@@ -9,6 +10,7 @@ include 'int-test-proxy-server'
 
 project(':xio').projectDir = "$rootDir/xio-core" as File
```
manimaul: i don't think these projectDir settings are even necessary

xio-zookeeper/build.gradle
```diff
@@ -0,0 +1,95 @@
+plugins {
+  id 'java'
```
manimaul: we should probably start using 'java-library'

xio-zookeeper/build.gradle
```diff
@@ -0,0 +1,95 @@
...
+  implementation(group: 'org.apache.curator', name: 'curator-recipes', version: curator_version) {
+    exclude(module: 'zookeeper')
+  }
+  implementation(group: 'org.apache.zookeeper', name: 'zookeeper', version: zookeeper_version) {
+    exclude(module: 'slf4j-log4j12')
+    exclude(module: 'log4j')
+  }
+  /* okhttp and moshi for JSON over HTTP */
+  implementation group: 'com.squareup.okhttp3', name: 'okhttp', version: okhttp_version
+  implementation group: 'com.squareup.moshi', name: 'moshi', version: moshi_version
```
manimaul: i like moshi but i think we're already using jackson (which is faster)

xio-zookeeper/build.gradle
```diff
@@ -0,0 +1,95 @@
+plugins {
+  id 'java'
+  id 'com.google.protobuf' version '0.8.5'
+  id 'com.github.kt3k.coveralls' version '2.8.2'
+  id 'jacoco'
+  id 'com.github.sherter.google-java-format' version '0.6'
+  id 'com.jfrog.bintray'
+}
+
+googleJavaFormat {
```
manimaul: goJF and jacoco configurations might/should be moved to the root level gradle

Issue #290, opened 2018-08-08
Allow for private key and certificates to be passed into TlsConfig


Farhie:
TlsConfig requires that your private key and certificate are on the classpath or a file. For instances where we want self signed certs we are generating them during build time, writing them out to a file and then overriding the config passed into the TlsConfig constructor. 

It'd be great if we were able to pass in our self signed cert and key.

jkinkead:
This in particular makes `TlsConfig` very difficult to use; but it also treats all configuration values as *required*, without providing fallback defaults. This makes it unsuitable for use as a utility outside of the `xio` toolchain.

What about switching `TlsConfig` to a builder, instead of an initialize-once bean class?

PR #305, opened 2018-09-14
Add a Builder for TlsConfig.

jkinkead:
Let most values of TlsConfig come from non-config-file sources.

This ended up going through a few iterations, but it seems reasonable now.

I got rid of some helpers for creating `TlsConfig` out of paths in typesafe configs, because it felt like a weird abstraction (and was only used in unit tests). I'm happy to add them back in.


jkinkead:
Fixes #290 .

xio-core/src/main/java/com/xjeffrose/xio/SSL/TlsConfig.java
```diff
@@ -172,33 +168,124 @@ private static ApplicationProtocolConfig buildAlpnConfig(Config config) {
...
+      ApplicationProtocolConfig alpnConfig,
+      List<String> ciphers,
+      ClientAuth clientAuth,
+      Boolean enableOcsp,
+      List<String> protocols,
+      Long sessionCacheSize,
+      Long sessionTimeout) {
+    // Validate that *all* parameters were set in the builder.
+    String errorMessage = "{} must be set in builder";
+    Preconditions.checkNotNull(useSsl, errorMessage, "useSsl");
```
manimaul: i'd expect lombok to be able to make this not null requirement check with an annotation

pdex: `lombok.Builder` is lacking in generated validation hooks, for that reason we've moved this sort of functionality over to google auto builder, we can refactor later.

PR #306, opened 2018-09-17
Address pr 303 comments

pdex:


xio-zookeeper/src/main/java/com/xjeffrose/xio/zookeeper/DeploymentCache.java
```diff
@@ -34,12 +30,14 @@ public void close() throws IOException {
   }
 
   public List<DeploymentPayload> getDeployments() throws IOException {
+    ObjectMapper objectMapper = new ObjectMapper();
```
manimaul: in general i tend to use an `injected` object mapper or just use a singleton
at least there is a single place where you'd add your adapters
this might effect performance (i'm not certain)

PR #307, opened 2018-09-18
dc/TACO-629 update the response code for when there are no new clients

khappucino:


khappucino:
Added HealthCheck config class so we can cross reference the NFE/NLP healthcheck routes and change them only in one place

PR #310, opened 2018-09-21
dc/TACO-597-update-healthcheckconfig adding port and tlsenabled in xio

khappucino:


xio-core/src/main/java/com/xjeffrose/xio/config/HealthCheckConfig.java
```diff
@@ -2,4 +2,7 @@
 
 public class HealthCheckConfig {
   public static final String healthCheckPath = "/health-check/";
+  public static final int healthCheckPort = 8000;
```
Rivukis: I believe the Java way of doing constants is to use all caps and underscores between words

PR #313, opened 2018-09-25
TACO-653 separate tls configuration into separate artifact

manimaul:
publishing xio-tls artifact for use in other components


xio-tls/src/main/java/com/xjeffrose/xio/tls/X509Certificate.java
```diff
@@ -14,36 +14,19 @@
  * limitations under the License.
```
pdex: I don't think anything is using this class? If so we can remove it and get rid of some javac warnings about touching `sun.security` classes.

xio-tls/src/main/java/com/xjeffrose/xio/tls/X509CertificateGenerator.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
```
pdex: pretty sure we can replace this with a light wrapper around this, maybe even just point users at this class:
https://netty.io/4.1/api/index.html?io/netty/handler/ssl/util/SelfSignedCertificate.html

PR #323, opened 2018-10-03
proxy bugfix - multiple backends - single client connection

manimaul:
fixing proxy when a single client connection attempts to reach multiple, desperate backends


int-test-proxy-server/src/test/java/com/xjeffrose/xio/proxy/ReverseProxyFunctionalTest.java
```diff
@@ -182,81 +213,126 @@ void post(int port, boolean sanity, Protocol expectedProtocol) throws Exception
...
     setupFrontBack(true, false);
 
-    get(port(), false, HTTP_2);
+    get(proxyPort(), HTTP_2, "/foo/", "/ifoo/", backEnd1);
+    assertEquals(1, backEnd1.getRequestCount());
+    assertProxiedRequests(1);
+  }
+
+  @Test
+  public void testMultipleBacksMixedGET() throws Exception {
```
manimaul: this is the failing test b4 the fix

PR #324, opened 2018-10-04
discard closed proxy clients in channel attributes

manimaul:
we want to dispose of proxy clients that are not "reusable" that have been stored in channel attributes just like we do for proxy clients in the pool

### DIFF
--- a/xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
+++ b/xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
@@ -153,6 +153,6 @@ public class Client {
    *     incorrect.
    */
   public boolean isReusable() {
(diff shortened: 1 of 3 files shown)


xio-core/src/main/java/com/xjeffrose/xio/http/Client.java
```diff
@@ -153,6 +153,6 @@ public ClientPayload(Request request, ChannelPromise promise) {
    *     incorrect.
    */
   public boolean isReusable() {
-    return manager.connectionState() == ClientConnectionState.CONNECTED;
+    return manager.connectionState() != ClientConnectionState.CLOSED_CONNECTION;
```
manimaul: we need to consider `connecting` and `not yet connected` in a reusable state

PR #325, opened 2018-10-04
proxy error response rather than conn close

manimaul:
responding with a 503 rather than closing connection to client when failing to write to a proxied backend


xio-core/src/main/java/com/xjeffrose/xio/http/ProxyHandler.java
```diff
@@ -171,12 +171,13 @@ private void writeClientRequest(ChannelHandlerContext ctx, Client client, Reques
             channelFuture.addListener(
                 (f) -> {
                   if (!f.isSuccess()) {
-                    // todo: (WK) do something more polite
-                    // we should probably emit a signal and have the application codec handle the event based
-                    // on the the response state
-                    // h2 - keep the connection open and close the stream
-                    // h1 - respond with 503 if not mid-stream
-                    ctx.close();
```
manimaul: @pdex - why did we close this ?

